PR #1, opened 2023-10-24
feat(platform-api): disable sites job for subscription cancel

abemscac:
WIP [#820](https://github.com/samatechtw/pubstudio/issues/820)


sampullman:
Thanks for updating the types here! I caught some migration issues and fixed them so at least format/lint/unit tests pass now.

PR #2, opened 2023-10-25
fix: editor invalid state

abemscac:
[#828](https://github.com/samatechtw/pubstudio/issues/828)

### DIFF
--- a/libs/frontend/feature-build-event/src/lib/use-build-event.ts
+++ b/libs/frontend/feature-build-event/src/lib/use-build-event.ts
@@ -72,8 +72,6 @@ export const useBuildEvent = () => {
   const { pressCopy, pressPaste } = useCopyPaste()
   const { pressDuplicate } = useDuplicateComponent()
   const { drag, stopDrag } = usePaddingMarginEdit()
-  const ed = editor.value
-  const context = site.value.context
   let buildWindow: HTMLElement | null = null
 
   const selectComponent = (component: IComponent) => {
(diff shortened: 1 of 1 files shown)


libs/frontend/feature-build-event/src/lib/use-build-event.ts
```diff
@@ -72,8 +72,6 @@ export const useBuildEvent = () => {
   const { pressCopy, pressPaste } = useCopyPaste()
   const { pressDuplicate } = useDuplicateComponent()
   const { drag, stopDrag } = usePaddingMarginEdit()
-  const ed = editor.value
-  const context = site.value.context
```
abemscac, line 76: Hi @sampullman 

It seems that this tiny little pattern is the source of the strange bug!
I think we've seen this before, but I couldn't find the related issue/comment from the repo now.
Even though all non-primitive values are being passed by reference in JS, the value (editor/site context in this case) will still have the old value after the site is replaced if we write it in this way. It's probably related to how Ref works in Vue?

I tried my best to search for the similar pattern in the whole workspace, luckily no other similar patterns are found. It seems that all of them are being wrapped inside a function/computed function, so they'll always have the latest value.

libs/frontend/feature-build/src/components/build-menu/BuildMenuFile.vue
```diff
@@ -8,7 +8,7 @@
         <Import></Import>
       </BuildMenuIconText>
     </SimpleFileUpload>
-    <BuildMenuIconText :text="t('templates')" @click="emit('showTemplates')">
+    <BuildMenuIconText class="templates" :text="t('templates')" @click="emit('showTemplates')">
```
abemscac, line 11: This is added for web E2E.

libs/frontend/feature-build-event/src/lib/use-build-event.ts
sampullman: Right! I should have caught this a while ago - if `useBuildEvent()` isn't called again, the old references will be stored. So if we want to use references like this to simplify the code, they would have to be `computed`. For example, `const context = computed(() => site.value.context)`. But that's not very useful!

PR #9, opened 2023-10-30
feat: behavior javascript syntax highlighting

abemscac:
[#746](https://github.com/samatechtw/pubstudio/issues/746)


libs/frontend/feature-build/src/lib/create-code-editor-view.ts
```diff
@@ -0,0 +1,67 @@
...
+        // FIXME
+        // 1. The current solution will clear the history because it creates a brand-new editor state.
+        //    Find a way to replace the content in the editor state while keeping everything.
+        // 2. New line characters are ignored if the text is copied from a source other than ProseMirror
+        //    editor. For example, VSCode.
+        newState = prosemirrorSetup({
+          content: `<pre><code>${textContent}</code></pre>`,
+          plugins: [...codeEditorPlugins],
+        })
+        newState.selection = Selection.atEnd(newState.doc)
```
abemscac, line 57: Hi @sampullman 

The "Maximum call stack exceeded" error is resolved by copying source code from [prosemirror-highlightjs](https://github.com/b-kelly/prosemirror-highlightjs) and commenting out the unused parts. But there're still some strange issues I didn't managed to solve 😵.

I left some comments here to describe the problem. What I don't understand is why these 2 problems only occurr when the copied text is from a source other than ProseMirror editor 🤔?
For example, if I try to copy some code from VSCode and paste it to the ProseMirror editor in BehaviorModal, these 2 problems will occurr. But if the code (text) is copied&pasted within the same editor, these problems will not show up!

libs/frontend/feature-build/src/lib/get-highlight-decorations.ts
```diff
@@ -0,0 +1,265 @@
+/**
+ * This file is a copy of "src/getHighlightDecorations.ts" from "prosemirror-highlightjs" library.
+ * @see https://github.com/b-kelly/prosemirror-highlightjs/blob/master/src/getHighlightDecorations.ts
+ */
```
abemscac, line 4: I ended up not installing `prosemirror-highlightjs` library and just copied the code to our workspace.

libs/frontend/feature-build/src/lib/prosemirror-highlighthjs-plugin.ts
```diff
@@ -0,0 +1,274 @@
...
+    props: {
+      decorations(this: Plugin<HighlightPluginState>, state) {
+        return this.getState(state)?.decorations
+      },
+    },
+    // `prosemirror-highlightjs` use this function to update node's language in a editor,
+    // which leads to a "Maximum call stack size exceeded" error. Since we only need to
+    // format JS code at the moment, we can just skip the following part for now.
+
+    // view(initialView: EditorView) {
```
abemscac, line 235: This `view` spec is what causes the "Maximum call stack size exceeded" error. From the code it seems that it's used to update the language of the code for highlightjs. But since we only care about JavaScript at the moment, I just commnted out the whole section to make it work.

libs/frontend/ui-theme/src/css/components/prosemirror-view.postcss
```diff
@@ -0,0 +1,66 @@
+/** This file is a copy of "node_modules/prosemirror-view/style/prosemirror.css" */
```
abemscac, line 1: This css file is coped from `node_modules/prosemirror-view/style/prosemirror.css`. This is necessary because it covers some of the style (i.e. `.ProseMirror pre`) in ProseMirror for code_block. Including these styles also solves the pre-wrap warning from ProseMirror.

libs/frontend/feature-build/src/lib/create-code-editor-view.ts
sampullman: I wasn't actually able to reproduce these two issues, unless I misunderstood something. Here's a recording of copy/paste from VSCode:


https://github.com/samatechtw/pubstudio-builder/assets/550982/7ec4d5d6-d890-4b5b-bc40-b5e1dc0cab8d

libs/frontend/feature-build/src/lib/prosemirror-highlighthjs-plugin.ts
sampullman: I guess it probably would have been sufficient to pass in a blank/empty `languageSetter` function when initializing the plugin, but disabling the code is clear.

sampullman:
We can open a new issue for the bug you mentioned, but I think it's fine to merge first!

libs/frontend/feature-build/src/lib/create-code-editor-view.ts
abemscac: Hi @sampullman 

I just tried it for a couple more times, and I created another issue #11 to keep track of the bugs!
I'll come back and try to fix it when I'm done with my current work!

sampullman: 🙏

PR #12, opened 2023-10-31
fix: various tooltip and ui fixes

sampullman:


apps/builder-demo/subset.config.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
       'prosemirror-state',
       'prosemirror-transform',
       'prosemirror-view',
+      'highlight.js/lib/core',
```
sampullman, line 35: Add new lib to subset for Docker build

libs/frontend/feature-build/src/components/toolbar/ToolbarBreakpointOption.vue
```diff
@@ -92,6 +97,7 @@ const breakpointHint = computed(() => {
     @mixin flex-center;
     width: 40px;
     height: 40px;
+    position: relative;
```
sampullman, line 100: @abemscac I'm not exactly sure whether it's due to updating `@floating-ui`, or something that happened when I migrated libs, but somehow all the tooltip positions were broken. I had to make sure all the parents had `position: relative;`, it makes sense but I'm not sure why it worked previously.

libs/frontend/ui-widgets/src/lib/InfoBubble.vue
```diff
@@ -73,20 +73,38 @@ const props = withDefaults(
...
-const show = ref(false)
 let hoverTimer: ReturnType<typeof setTimeout> | undefined
 
 const clickInfo = () => {
-  show.value = !show.value
+  setShow(!show.value)
   cancelBubbleDelay()
 }
 
+const setShow = (newShow: boolean) => {
```
sampullman, line 97: Rewrote all consumers of `useTooltip` to use `show` provided by the composable, instead of trying to track separately. `tooltipMouseEnter` should probably be renamed, since it's now used generally to indicate the tooltip should be shown.

libs/frontend/ui-widgets/src/lib/form/PSMultiselect.vue
```diff
@@ -152,6 +152,7 @@ const {
 } = useDropdown({
   clickawayIgnoreSelector: `div[data-toggle-id="${dataToggleId}"]`,
   offset: { mainAxis: 0, crossAxis: 0 },
+  layoutShift: false,
```
sampullman, line 155: @abemscac I noticed some dropdown menus will repeatedly receive the `autoUpdate` callback from `@floating-ui`. Once the menu is clicked, it will settle and only be called as expected. Setting `autoUpdate` option `layoutShift` to false fixes this, although it means the menus won't automatically shift to the top/bottom on scroll.

For now this is ok, since I don't really like the automatic menu shifting anyway, but it would be nice to figure out the autoUpdate issue. It might be related to the flex layout interaction with `IntersectionObserver` in floating-ui, but I didn't figure out the exact reason.

libs/frontend/util-tooltip/src/lib/use-tooltip.ts
```diff
@@ -45,51 +55,94 @@ export const useTooltip = (options?: IUseTooltipOptions): IUseTooltip => {
...
     position: strategy.value,
     top: `${y.value ?? 0}px`,
     left: `${x.value ?? 0}px`,
     width: 'max-content',
     'z-index': 1001,
-  }))
+  })
+
+  const updatePosition = async () => {
+    if (itemRef.value && tooltipRef.value) {
```
sampullman, line 110: @abemscac I rewrote `use-tooltip` to make use of `autoUpdate`, because I couldn't figure out why all the tooltips were suddenly placed incorrectly. For some reason, `useFloating` resulted in top/left = 0, where before it worked correctly.

This more closely matches floating-ui examples for arrow styling, so I think it's ok, even though I don't understand how the previous code broke.

PR #14, opened 2023-11-02
feat(platform-api): api template categories #848

ClaireMao1996:
Close https://github.com/samatechtw/pubstudio/issues/848


sampullman:
Thanks! I'll fix the frontend typing issues. I wonder if it's worth moving the platform API types back to the platform monorepo 🤔

ClaireMao1996:
Oh, thank you! I should aware that. 

> I wonder if it's worth moving the platform API types back to the platform monorepo 

But are some of these also shared in `pubstudio-builder`?

PR #22, opened 2023-11-06
feat(site-api): visiting the preview page before editor saves is conf…

ClaireMao1996:
Close https://github.com/samatechtw/pubstudio/issues/857

### DIFF
--- a/apps/api-site-test/test/integration/get-site.spec.ts
+++ b/apps/api-site-test/test/integration/get-site.spec.ts
@@ -51,6 +51,44 @@ describe('Get Site', () => {
       expect(body.pages).toBeDefined()
     })
 
+    it('returns site data if current update_key does not match the saved one', async () => {
+      const response = await api
+        .get(`${testEndpoint}/${siteId}`)
+        .query({ update_key: 1 })
+        .set('Authorization', adminAuth)
+        .expect(200)
+      const body: IGetSiteApiResponse = response.body
+
+      expect(body.id).toEqual(1)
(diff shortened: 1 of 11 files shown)


ClaireMao1996:
I suddenly thought, can we use `updated_at` as the update_key?

sampullman:
Sorry I missed your comment! You're right, I'm not sure why I didn't think of that.

We'll need something more advanced in the future if we support multiple people editing at the same time, but for now I don't see why `updated_at` can't be used.

ClaireMao1996:
If we decide to use `updated_at` as the update key, I think backend may not need to changes? However, I'm not entirely sure if this approach will be sufficient for our future needs.

sampullman:
The backend should still be updated to return no data when `updated_at` in the request matches, is that what you mean?

ClaireMao1996:
I mean, if we use `updated_at` as the update key, `updated_at` will be updated every time when the site is updated. So, when the frontend is on the preview page and polls the API every few seconds, it can use the `updated_at` value from the response to compare it with the current `updated_at` to determine if there's a need to refresh the entire page. Does that make sense, or have I misunderstood something? 🤔

sampullman:
That's correct! I just meant that the `get-site` endpoint still has to be adjusted only return data if `updated_at` doesn't match, or if it's not provided in the query.

Otherwise it seems like we don't need to modify the `update-site` endpoint or DB entities.

ClaireMao1996:
Oh, I see! Thank you! I think I'll create a new PR for this. Adjust to the version with 'updated_at'

PR #23, opened 2023-11-06
feat: workaround for editor root 100% height

sampullman:


apps/web-site/src/app/App.vue
```diff
@@ -164,8 +164,6 @@ body {
   margin: 0;
   width: 100%;
   height: 100%;
-  background-color: #f8f5f0;
```
sampullman, line 167: I had to remove these for the workaround to take effect.

libs/frontend/feature-build/src/components/BuildContent.vue
```diff
@@ -58,10 +63,25 @@ const innerStyle = computed<StyleValue>(() => {
 
   const innerHeight = buildContentWindowHeight / builderScale
 
+  // Set builder background/color to page root to handle the case
+  //   when child components overflow the root
+  let rootStyles: IRawStyle | undefined = undefined
+  if (activePage.value) {
```
sampullman, line 69: @abemscac This is the workaround for the editor. It's probably not very efficient to do it this way, but it seems to work for all cases.

libs/frontend/feature-render/src/lib/get-page-style.ts
```diff
@@ -46,6 +48,34 @@ export const getBuildPageStyle = (site: ISite, page: IPage): IRawStyleRecord =>
   return pageStyle
 }
 
+export const getRootBackgroundStyle = (
```
sampullman, line 51: @abemscac This is messy, maybe there's a better way to do it. It just grabs the `background` and `background-color` properties from the page root for all breakpoints.

libs/frontend/feature-render/src/lib/use-render.ts
```diff
@@ -62,8 +65,11 @@ export const useRender = (options: IUseRenderOptions): IUseRender => {
   const livePageComponentStyle = computed(() => {
     let styleContent = ''
     if (activePage.value && site.value) {
+      const htmlStyle = getRootBackgroundStyle(site.value.context, activePage.value)
       const pageStyle = getLivePageStyle(site.value.context, activePage.value)
-      styleContent = queryStyleToString(site.value.context, pageStyle)
+      styleContent =
+        queryStyleToString(site.value.context, htmlStyle) +
```
sampullman, line 71: @abemscac The `html` style for preview and live mode is added here.

sampullman:
@abemscac I'll wait until you have a chance to take a look at this tomorrow before merging! I tested a few scenarios and it seems to work, but I could have missed something.

abemscac:
Hi @sampullman 

The workaround seems to work well! The only thing I've noticed (it's probably not an issue at all) is that if `height: 100%` is applied to the root component, the height of the root component will be 100% of editor-view in build mode, so we'll see the bottom border of the root component when we scroll down a little bit:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/9fc628c9-ef2b-4d00-8213-463f5dca5cb8

This means that if `overflow: auto` is applied to the root component, two scrollbars are expected to be seen on the screen. If it's `overflow: hidden` then the content after 100vh will become invisible.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/8855caf6-93a2-4fc5-8ad5-2c59873354e6

But I don't think this will be a problem, because for most cases people will not add `overflow` to the root component. That being said, do you think it's necessary to remove `height: 100%` from the default style of the root component? The fix still seemed to be working after  it's removed.

sampullman:
Good point! I will remove `height: 100%` from the root. I think it's only used for a few types of sites, usually a template can take care of those details if needed.

sampullman:
@abemscac After some further testing I decided to keep `height: 100%` on the root component for now. It's better for the situation where the user sets child component height to 100%, but worse if they set the root overflow to hidden (as you mention).

Maybe we can think of a solution that's best for all situations later! Some options could be automatically setting certain root properties when the user changes something (e.g. remove 100% height if overflow is set on root), or providing some explanation about strategies for handling root styling.

PR #24, opened 2023-11-07
fix: cannot drop component above container in component tree

abemscac:
[#854](https://github.com/samatechtw/pubstudio/issues/854)
[#864](https://github.com/samatechtw/pubstudio/issues/864)

### DIFF
--- a/libs/frontend/feature-build/src/components/ComponentTreeItem.vue
+++ b/libs/frontend/feature-build/src/components/ComponentTreeItem.vue
@@ -1,6 +1,7 @@
 <template>
   <div
     :id="treeItemId"
+    ref="elementRef"
     class="component-tree-item"
     :class="{
       dragging: dndState?.dragging,
@@ -16,7 +17,6 @@
     @drop.stop="drop"
   >
     <div
-      ref="elementRef"
       class="component-title"
       :class="{
         selected: currentSelected,
--- a/libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
+++ b/libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
@@ -133,6 +133,7 @@ export const BuilderDndComponent = defineComponent({
         getParentId: () => dragCmp?.parent?.id,
         getComponentIndex: () => componentIndex,
         isParent: component.value.id !== dragCmp?.id,
+        stopAtFirstMatch: true,
       })
     })
     onMounted(() => {
@@ -167,7 +168,9 @@ export const BuilderDndComponent = defineComponent({
               onDrop: dnd.drop,
               onDragend: dnd.dragend,
             }
-          : undefined
+          : {
+              onDragend: dnd.dragend,
+            }
 
       const props = {
         ...propsContent.props,
--- a/libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
+++ b/libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
@@ -133,7 +133,6 @@ export const BuilderDndComponent = defineComponent({
         getParentId: () => dragCmp?.parent?.id,
         getComponentIndex: () => componentIndex,
         isParent: component.value.id !== dragCmp?.id,
-        stopAtFirstMatch: true,
       })
     })
     onMounted(() => {
--- a/libs/frontend/feature-build/src/lib/dnd/drag-drop.ts
+++ b/libs/frontend/feature-build/src/lib/dnd/drag-drop.ts
@@ -13,6 +13,18 @@ export interface IDropResult {
   to: IComponentPosition
 }
 
+export interface IOnDragOptions {
+  e: DragEvent
+  context: ISiteContext
+  dragSrc: IDraggedComponent
+  hoverCmpIndex: number
+  hoverCmp: IComponent
+  elementRef: HTMLElement | undefined
+  // Only check for top&bottom hover
+  verticalOnly?: boolean
+  stopAtFirstMatch?: boolean
+}
+
 const getMousePosition = (e: DragEvent): XYCoord => {
   return { x: e.clientX, y: e.clientY }
 }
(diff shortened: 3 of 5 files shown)


libs/frontend/feature-build/src/components/ComponentTreeItem.vue
```diff
@@ -16,7 +17,6 @@
     @drop.stop="drop"
   >
     <div
-      ref="elementRef"
       class="component-title"
```
abemscac, line 20: Hi @sampullman 

I move the `elementRef` of component-tree-item from title element to the whole container (title + children) so that the behavior feels more natural when the bottom of the title element is hovered.

Before:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/15e85312-bbaa-44dd-840e-dc975be3789f

After:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/a7c5d79e-9f77-45c9-9912-566c22cc0baf

The bottom border of a component-tree-item is expected to show when the bottom section of the whole container is hovered, instead of when the bottom section of the title element is hovered.

libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
```diff
@@ -167,7 +167,9 @@ export const BuilderDndComponent = defineComponent({
               onDrop: dnd.drop,
               onDragend: dnd.dragend,
             }
-          : undefined
+          : {
+              onDragend: dnd.dragend,
+            }
```
abemscac, line 172: This is for [#854](https://github.com/samatechtw/pubstudio/issues/854).

The cause of this bug is that even though we forbid a component from being dragged when its ProseMirror editor is active, its parent can still be dragged at the same time. Since a ProseMirror editor is still active when the parent is dropped, the `onDragend` event will not be attached to the component, leading to DND state not being cleared on drop.

So in short, we should attach `onDragend` event to DND component all the time so that DND state can be cleared on drop.

libs/frontend/feature-build/src/lib/dnd/drag-drop.ts
```diff
@@ -41,16 +52,10 @@ const checkSelfDrop = (
...
-  e: DragEvent,
-  context: ISiteContext,
-  dragSrc: IDraggedComponent,
-  targetIndex: number,
-  target: IComponent,
-  elementRef: HTMLElement | undefined,
-  // Only check for top&bottom hover
-  verticalOnly?: boolean,
-): IDropProps => {
+export const onDrag = (options: IOnDragOptions): IDropProps => {
```
abemscac, line 55: For [#864](https://github.com/samatechtw/pubstudio/issues/864), I noticed that the drop actually works under some circumstances (in some scenarios it really didn't work!), it's just that we have to move the cursor slightly to the side.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/ee960b22-607b-4f6b-99ee-2177dc9e0aec

This happens because we check for hoverSelf first, then top/bottom and left/right. The easiest solution I could think of is to rearrange the order to make it check for top/bottom and left/right first, then hoverSelf.

I also made some small changes to the check condition so that dropping a component to the same level as its parent can work properly. It seems to work now, feel free to let me know if you notice any problem!

libs/frontend/feature-build/src/lib/dnd/row-layout.ts
```diff
@@ -18,7 +18,7 @@ interface IHandleHoverProps {
   targetMiddle: XYCoord
   haveSameParent: boolean
   dragComponentIndex: number
-  targetIndex: number
+  hoverCmpIndex: number
```
abemscac, line 21: The changes made to this file are just refactoring (prop renaming, destructing assignment, etc), the logic stays exactly the same.

abemscac:
Hi @sampullman 

I think this PR is ready for review, thank you!

libs/frontend/feature-build/src/components/ComponentTreeItem.vue
sampullman: Nice!

libs/frontend/feature-build/src/lib/dnd/drag-drop.ts
sampullman: Makes sense to me 👍

sampullman:
I think the related issues are fixed! But I noticed another issue - I don't think the black line should show above or below the root component, since we can't drop things there. Also, I noticed it's sometimes not possible to drop on a container (e.g. show the red outline), maybe when there are children? I'm not sure that's a problem though.

If you prefer, I can create a new issue and merge this first!


https://github.com/samatechtw/pubstudio-builder/assets/550982/f3d0f786-9b4f-44cf-8a69-be8e68eb9a71

abemscac:
I see! Sorry for creating even more bugs 😬.
I'll try to solve them all at once within the same PR!

sampullman:
Thanks, no worries!

libs/frontend/feature-build/src/lib/dnd/row-layout.ts
```diff
@@ -38,74 +38,90 @@ export const isRowLayout = (
...
 
-export const handleRowLayoutHover = ({
-  mousePosition,
-  targetMiddle,
-  haveSameParent,
-  dragComponentIndex,
-  targetIndex,
-  dropProps,
-}: IHandleHoverProps) => {
+export const handleRowLayoutHover = (props: IHandleHoverProps) => {
```
abemscac, line 41: Hi @sampullman 

I adjusted the handler of hover handler here to fix the new bug you found.
Previously we check if the cursor is at the top/bottom or left/right side of the middle of the element, and stop checking for self hover if any one of them is true.
After this adjustment, only hovering on the edge (top/bottom or left/right) will trigger the same logic.

libs/frontend/feature-build/src/lib/dnd/row-layout.ts
```diff
@@ -38,74 +38,90 @@ export const isRowLayout = (
...
   const hoverBottom =
-    mousePosition.y > targetMiddle.y &&
-    (!haveSameParent || dragComponentIndex !== targetIndex + 1)
+    mousePosition.y <= bottom &&
+    mousePosition.y >= bottom - hoverableHeight &&
+    (!haveSameParent || dragComponentIndex !== hoverCmpIndex + 1)
 
   dropProps.hoverTop = hoverTop
   dropProps.hoverBottom = hoverBottom
-  dropProps.destinationIndex = targetIndex
```
abemscac, line 95: This strange adjustment is intended to fix a new bug where the component will be moved up one level even if it's not being dropped on the top side of its parent.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/4bdf764a-d213-4ea4-9435-e2a6c736b6bf

sampullman:
Looks good to me, nice work!

PR #25, opened 2023-11-08
feat(site-api): get site add update key query

ClaireMao1996:
Close https://github.com/samatechtw/pubstudio/issues/857

### DIFF
--- a/backend/lib-shared-types/src/dto/site_api/get_site_dto.rs
+++ b/backend/lib-shared-types/src/dto/site_api/get_site_dto.rs
@@ -1,7 +1,6 @@
-use chrono::{DateTime, Utc};
-use serde::{Serialize, Deserialize};
+use serde::{Deserialize, Serialize};
 
-use crate::entity::site_api::site_entity::SiteEntity;
+use crate::{entity::site_api::site_entity::SiteEntity, shared::js_date::JsDate};
 
 #[derive(Serialize)]
 pub struct GetSiteDto {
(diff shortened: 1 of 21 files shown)


backend/lib-shared-types/src/dto/site_api/get_site_dto.rs
```diff
@@ -19,6 +19,30 @@ pub struct GetSiteDto {
     pub disabled: Option<bool>,
 }
 
+#[derive(Deserialize)]
+pub struct GetSiteQuery {
+    pub update_key: Option<String>,
```
ClaireMao1996, line 24: Frontend only needs to take the `updated_at` received in the response, such as `2023-11-08T02:39:36Z` and use it as the update key

ClaireMao1996:
Hi @sampullman ,

This PR is ready for review, sorry for the delay!

backend/site-api/src/app/site/get_site.rs
```diff
@@ -78,11 +94,32 @@ pub async fn get_site(
...
-        )));
+
+        if let Some(update_key) = query.update_key {
+            return DateTime::parse_from_rfc3339(update_key.as_str())
+                .map_err(|e| {
+                    ApiError::internal_error().message(format!("Date parsing error: {}", e))
+                })
+                .and_then(|date_time| {
+                    if site.updated_at == date_time {
+                        Ok(StatusCode::NO_CONTENT.into_response())
```
sampullman, line 105: I adjusted the endpoints to return `204` response instead of empty data. The reason is that it's a little simpler for the frontend to deal with, and slightly easier to maintain in the backend, if site response fields change. I had to make this `get_site_result` helper function in order to return `Result<Response, ApiError>`, since Axum won't allow that to be returned from the top level handler.

sampullman:
It seems to work well! I am polling every 3 seconds on the frontend, but only when the tab is visible. I also check for new data when the tab goes from hidden -> visible.

It might be better to reduce the polling time, but we can always adjust it later.

PR #26, opened 2023-11-08
fix: hash links don't work in preview page

abemscac:
[#872](https://github.com/samatechtw/pubstudio/issues/872)


libs/frontend/feature-preview/src/lib/preview-component.ts
```diff
@@ -64,21 +66,45 @@ export const PreviewComponent = () => {
...
-            const { search, hash } = window.location
             const copiedRenderProps = {
               ...renderProps,
-              to: `${hrefProp}${search}${hash}`,
+              to: `${previewPathName}${queryString}${userHash || editorHash}`,
             }
             delete (copiedRenderProps as Record<string, unknown>).href
 
-            return h(RouterLink, copiedRenderProps, children)
+            return h(RouterLink, copiedRenderProps, () => children)
```
abemscac, line 107: I noticed that there's a warning says the default slot content of RouterLink should be a function (both in preview and live mode), so I change this (and the one in `live-comonent.ts`) from `children` to `() => children`, which seems to fix the issue.

Screenshot 2023-11-07 at 1 57 16 PM

libs/frontend/util-router/src/lib/scroll-behavior.ts
```diff
@@ -87,15 +71,14 @@ export function scrollToPosition(
...
       return
     }
-    scrollToOptions = getElementPosition(el, position)
-  } else {
-    scrollToOptions = position
-  }
 
-  if (scrollRoot) {
-    scrollRoot.scroll(scrollToOptions)
+    el.scrollIntoView()
```
abemscac, line 75: Hi @sampullman 

I noticed that the scroll behavior of a hash link in live mode is a little strange; the position seems to jump/switch between two places upon click.
Since this `if` section is intended to scroll to a specific element, `scrollIntoView()` seems to work well here. I may have missed something, it would be great if you could double check if this makes sense!

https://github.com/samatechtw/pubstudio-builder/assets/59676941/a8f61ef7-886e-4847-b02c-74dbdab13ff0

abemscac:
Hi @sampullman 

I think the hash link is working now, but for some reason there will be a significant delay between the click and scroll, and I just couldn't figure out why 🤔.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/e9be2647-3589-4478-b41c-e055fdc55b47

The RouterLink being used in preview mode is from `vue-router`, but I don't think I've seen this kind of delay with `vue-router` before. I thought maybe it's because there are too many elements with `id` in the document, but the delay still happens even if there are only two elements in the document. Do you have any clue what might be the cause of this problem?

libs/frontend/feature-preview/src/lib/preview-component.ts
sampullman: Oh yeah I remember seeing that! I think it's a strange warning though, I wonder if it's really more efficient.

sampullman:
I'll try to look into the delay issue when I review it! I can't recall seeing that issue before. We do have this code in `router.ts` to delay scroll, but that should only be on initial page load, so I don't think it's related:

```ts
const router = createRouter({
  history: createWebHistory(),
  scrollBehavior(to, from, savedPosition) {
    if (to.hash) {
      return new Promise((resolve, _reject) => {
        setTimeout(() => {
          resolve({ el: to.hash })
        }, 500)
      })
    }
```

libs/frontend/feature-preview/src/lib/merge-search.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * searchB will take precedence over searchA.
+ */
+export const mergeSearch = (searchA: string, searchB: string): URLSearchParams => {
+  const urlA = new URL(`https://tmp${searchA}`)
```
sampullman, line 5: Is it necessary to use `tmp` here? I also wasn't sure about the `tmp` in `href-to-url.ts`

libs/frontend/util-router/src/lib/scroll-behavior.ts
```diff
@@ -20,20 +20,6 @@ export interface ScrollPositionElement extends ScrollToOptions {
...
 
-function getElementPosition(
-  el: Element,
-  offset: ScrollPositionCoordinates,
-): _ScrollPositionNormalized {
-  const docRect = document.documentElement.getBoundingClientRect()
-  const elRect = el.getBoundingClientRect()
-
-  return {
-    behavior: offset.behavior,
```
sampullman, line 31: Won't removing this function break the situation where `behavior` has been set? Maybe I'm missing something.

libs/frontend/feature-preview/src/lib/merge-search.ts
abemscac: The intention of `tmp` here is to make it a valid argument for the `URL` class. Anything that does not start with `http(s)://{host}` will lead to an error says "Uncaught TypeError: Failed to construct 'URL': Invalid URL".

Originally I tried to use `window.location.host` so that we don't have to use `tmp`, but then I realized that the test would fail in that case because `window` only exists in browser environment.

The use of `URL` class is to save us some time parsing the hash and query params ourselves. I should've left a comment here. Sorry about that!

libs/frontend/util-router/src/lib/scroll-behavior.ts
abemscac: Oh, I removed it because I only found one occurrence of `getElementPosition` in the workspace, and that usage is being replaced by `element.scrollIntoView()` by me in this PR.
Is it actually a function that gets called somewhere else?

sampullman: No, I was just curious if the new way of using scrollIntoView covers all the same usecases of getElementPosition. It should be fine!

PR #32, opened 2023-11-10
fix: image with non-zero margin shifts on hover

abemscac:
[#860](https://github.com/samatechtw/pubstudio/issues/860)

### DIFF
--- a/libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
+++ b/libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
@@ -37,13 +37,16 @@ const renderImageHover = (
   componentClass: string[],
   content: IContent,
 ): VNode => {
-  const img = h(component.tag as string, props)
   const imgElement = document.getElementById(component.id)
 
   let [width, height] = ['', '']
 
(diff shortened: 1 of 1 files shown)


libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
```diff
@@ -79,6 +82,16 @@ const renderImageHover = (
...
 
+  // Pass the margin of img to the wrapper element, and set the margin of the rendered img
+  // to 0 during hover to avoid shifting issue.
+  const imgStyleProp = structuredClone(props.style ?? {}) as Record<string, unknown>
+  imgStyleProp['margin'] = '0'
+
+  const img = h(component.tag as string, {
+    ...props,
+    style: imgStyleProp,
+  })
```
abemscac, line 93: Hi @sampullman 

It seems that only `margin` would lead to the shift issue we've observed in the editor, while `padding` will not.
So I only pass the `margin` of `img` to the wrapper element here, which seems to fix the issue.

sampullman: Got it!

PR #33, opened 2023-11-10
fix: text gradient color missing in prosemirror editor

abemscac:
[ #874](https://github.com/samatechtw/pubstudio/issues/874)

### DIFF
--- a/libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
+++ b/libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
@@ -4,10 +4,18 @@
 
 <script lang="ts" setup>
 import { computed, onMounted, onBeforeUnmount, ref, toRefs, watch } from 'vue'
-import { IComponent, IEditorContext } from '@pubstudio/shared/type-site'
+import { Css, IComponent, IEditorContext } from '@pubstudio/shared/type-site'
 import { getProseMirrorContainerId } from '@pubstudio/frontend/feature-editor'
 import { prosemirrorSetup } from '@pubstudio/frontend/util-edit-text'
+import { findStyles } from '@pubstudio/frontend/util-component'
+import {
+  activeBreakpoint,
+  descSortedBreakpoints,
+} from '@pubstudio/frontend/feature-site-source'
+import { isTextGradient } from '@pubstudio/frontend/util-gradient'
+import { Plugin } from 'prosemirror-state'
 import { createComponentEditorView } from '../../lib/create-editor-view'
+import { useBuild } from '../../lib/use-build'
 
 const props = defineProps<{
   component: IComponent
@@ -15,6 +23,7 @@ const props = defineProps<{
 }>()
 
 const { component, editor } = toRefs(props)
+const { site } = useBuild()
 
 const container = ref<HTMLDivElement>()
 const containerId = computed(() => getProseMirrorContainerId(component.value))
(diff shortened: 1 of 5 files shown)


libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
```diff
@@ -25,20 +34,66 @@ const mountProseMirrorEditor = (
...
       {
         content: selectedComponent.content ?? '',
+        plugins,
       },
       container,
     )
   }
 }
 
+const gradientColorStyle = computed(() => {
```
sampullman, line 59: This seems like a pretty complex change just to get gradients working. Will it apply more generally if we also need to pass other styles to the prosemirror editor?

sampullman:
It works well, thanks!

libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
abemscac: Oh, I was just about to leave some comments to describe what I've changed, but you reviewed it so fast that I'm not even done 😅.

I accidentally found out a couple minutes ago that if we remove the `position: relative` at `pubstudio-builder/libs/frontend/ui-theme/src/css/components/prosemirror-view.postcss:4`, the text gradient problem would be magically solved!
I'm not very sure why ProseMirror set that style; I guess it's because ProseMirror editor is allowed to display HTML content, the `position: relative` is to make sure that absolute positioning elements will not be at somewhere outside the editor.

If we plan to allow users to paste HTML content to the editor in the future, maybe we should keep this `position: relative`. But that means we'll have to think of a way to create PubStudio components based on the pasted HTML content.

libs/frontend/feature-build/src/components/toolbar/ToolbarText.vue
```diff
@@ -251,16 +230,7 @@ const iconColor = computed(() => {
...
-  () => [editor?.value?.selectedComponent, editor.value?.cssPseudoClass],
-  () => {
-    const size = getStyleValue(Css.FontSize)
-    fontSize.value = size?.replace('px', '') ?? ''
-  },
-  {
-    immediate: true,
-  },
-)
+const { fontSizePsInputRef, fontSize, fontSizeSuffix, setFontSize } = useToolbarFontSize()
```
abemscac, line 233: This change is intended to solve a warning that occasionally shows up (when a component with non-px font-size is selected).

Screenshot 2023-11-10 at 10 54 36 AM

Turns out it was because we've only handle px in the font-size input in the toolbar, but the font-size of a component can actually be %, em, rem, etc. I create a new composable `useToolbarFontSize` for this. and move all font-size related code from `ToolbarText.vue` to the new composable.

libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
sampullman: Ah, sorry about that! I was going through all the PRs just now.

So do you prefer to revert this change and remove `position: relative` instead? I don't really understand why that fixes it though, and I worry a bit that it will be a fragile solution.

Long term, my idea is that a subset of html can be copy/pasted into the editor, maybe only `span/h/a/p`, but I'm not sure yet. ProseMirror would still control the html/layout, but we would have a plugin to allow styling text selections based on editor commands. I think the styles would be inlined in that case.

libs/frontend/feature-build/src/components/toolbar/ToolbarText.vue
sampullman: Got it!

libs/frontend/feature-build/src/components/toolbar/ToolbarText.vue
```diff
@@ -271,13 +241,16 @@ watch(
 }
 
 .font-size {
-  width: 72px;
+  width: 76px;
```
abemscac, line 244: The text in the font-size input seems a little crowded when the unit is rem instead of px, so I added a little more space to it.

libs/frontend/feature-build/src/components/toolbar/ToolbarText.vue
```diff
@@ -271,13 +241,16 @@ watch(
...
   margin-left: 4px;
   :deep(.ps-input) {
     border: 1px solid $color-light1;
     background-color: rgba(255, 255, 255, 0.8);
     font-size: 15px;
     padding-right: 32px;
   }
+  :deep(.ps-input-suffix) {
+    user-select: none;
+  }
```
abemscac, line 254: I add `user-select: none` to the suffix of font-size input to implicitly inform users they (currently) can't change the unit of font-size in the toolbar.

Before:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/5a7db745-99ca-45aa-b34e-3148b1aef03b

After:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/787d589a-6386-464b-9371-2f803f79fd33

I was thinking if we should use a dropdown of font-size units to replace the suffix text so that users can change not only the size but the unit at the same time, but I think that'll take some time; maybe we can come back here when most of the bugs are fixed! Also, I'm not sure if we should allow users to change the unit if font-size is actually a resolved value of a theme variable (e.g. `${size-h1}`).

I hope this change can temporarily let users giving up the idea of changing font-size unit in the toolbar 😬.

libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
abemscac: I think the current solution (passing styles from component to ProseMirror editor) is more reliable, because I also don't understand why removing `position: relative` fixes the bug 😅! I'm not sure if it's something related to `-webkit-text-fill-color` and/or `-webkit-background-clip`.

From [this sandbox](https://stackblitz.com/edit/web-platform-gfdp58?file=index.html) it seems that text gradient can be applied to the descendant no matter how deep it is. But adding `position: relative` in the middle does make the text invisible, just like what we've seen in the ProseMirror editor. So I think it's a HTML thing rather than a ProseMirror-specific issue.

sampullman: Interesting! Let's stick with your solution.

Although I just noticed something strange, which may not be fixable:

Screenshot 2023-11-10 at 3 57 10 PM

The gradient is applied over all the text, rather than each individual line. I suppose it would be really rare to actually want to have multiple lines of gradient text though 🤔

libs/frontend/feature-build/src/components/toolbar/ToolbarText.vue
sampullman: Works for me! I made a note to improve font-size unit handling later.

PR #58, opened 2023-11-14
feat: scroll to affeced component on undo/redo

### LINKED ISSUE #19: Scroll to affected component on undo/redo
When undo/redo, scroll to&regain focus on the effected component. Sometimes the affected component is in another page, but users may not know this, so they'll think that undo is broken because the styles/content of components are not changing.

Undo/redo should automatically switch pages and/or scroll to the affected component. This functionality probably belongs in `useHistory`, which already has a dependency on `useBuild`. Some updates to command handling may be necessary in order to figure out which component a command affects.

The general concept of scrolling to a component can be applied when selecting from the component tree as well.

abemscac:
Close #19

### DIFF
--- a/libs/frontend/feature-command/src/lib/component/remove-component.spec.ts
+++ b/libs/frontend/feature-command/src/lib/component/remove-component.spec.ts
@@ -29,6 +29,10 @@ describe('Remove Component', () => {
   let componentCount: number
 
   beforeEach(() => {
+    // Select the component that will be deleted, so undo matches
+    // Undoing RemoveComponent selects the un-deleted component, which may not be
+    // the component that was previously selected.
+    mockSerializedSite.editor!.selectedComponentId = 'test-c-1'
     siteString = JSON.stringify(mockSerializedSite)
     site = deserializeSite(siteString) as ISite
     componentCount = Object.keys(site.context.components).length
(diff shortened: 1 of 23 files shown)


libs/frontend/feature-command/src/lib/page/add-page.ts
```diff
@@ -42,15 +43,16 @@ export const undoAddPage = (site: ISite, data: IAddPageData) => {
...
   if (componentTreeExpandedItems) {
     delete componentTreeExpandedItems[rootId]
   }
+
   // Trigger page change event
   triggerEditorEvent(site, EditorEventName.OnPageChange)
+
+  // Set active page and selected component
+  setActivePage(site.editor, activePageRoute)
+  setSelectedComponent(site, prevSelectedComponent)
```
abemscac, line 57: Hi @sampullman 

The scroll to affected component feature was added to the `setSelectedComponent` helper, so commands where `setSelectedComponent` is used will automatically scroll to the affected component when applied.
But since commands may also be applied when `triggerEditorEvent` is called, I have to move `setActivePage` and `setSelectedComponent` down so that we can select on the correct component.
Although this simple solution worked, the downside is that users might see a quick flash between components/pages during redo/undo.

sampullman: Thanks, I'll keep that in mind! I think it's fine.

sampullman:
Overall it seems to work well, nice work! I adjusted the unit tests, previously we could assume that undoing a command makes the site data exactly match the previous version before the original command. Now, it is possible that the selected component will be different before and after a command/undo sequence. I don't think it's a problem, but something to be aware of with the unit tests.

libs/frontend/feature-command/src/lib/component/remove-component.spec.ts
```diff
@@ -29,6 +29,10 @@ describe('Remove Component', () => {
   let componentCount: number
 
   beforeEach(() => {
+    // Select the component that will be deleted, so undo matches
```
sampullman, line 32: Fix tests and add explainer comment

PR #59, opened 2023-11-14
refactor: skip slack logging in dev/ci

sampullman:


backend/lib-shared-site-api/src/layer/slack_layer.rs
```diff
@@ -200,9 +204,16 @@ where
         let blocks_result = Self::build_blocks(message, event, metadata);
         match blocks_result {
             Ok(blocks) => {
+                if self.exec_env == ExecEnv::Dev || self.exec_env == ExecEnv::Ci {
```
sampullman, line 207: @ClaireMao1996 I disabled slack logging in dev/ci, and printed directly to console. I tried to do it in `worker.rs`, but ran in to some strange problems with the channel closing.

backend/lib-shared-site-api/src/layer/slack_layer.rs
```diff
@@ -200,9 +204,16 @@ where
...
+                    println!("Skip slack log: {blocks}");
+                    return;
+                }
                 let payload = SlackPayload::new(blocks, webhook_url.to_string());
                 if let Err(e) = self.slack_sender.send(WorkerMessage::Data(payload)) {
-                    tracing::error!(err = %e, "failed to send slack payload to given channel")
+                    println!(
```
sampullman, line 213: Change this to `println` because it's possible to get in an infinite loop of logging errors.

PR #63, opened 2023-11-16
feat: font previous in theme menu and style toolbar #49

### LINKED ISSUE #49: Font previews in theme menu and style toolbar
Show font preview when adding fonts in theme menu, and when selecting fonts from the font-family dropdown in the style toolbar. 

Apply those fonts to the options in the menu so that font names are displayed in their own font.

sampullman:
Close #49

### DIFF
--- a/libs/frontend/ui-widgets/src/lib/form/PSMultiselect.vue
+++ b/libs/frontend/ui-widgets/src/lib/form/PSMultiselect.vue
@@ -43,7 +43,10 @@
           class="ms-item"
           @click="select(index)"
         >
-          {{ l }}
+          <slot v-if="customLabel" :label="l"></slot>
+          <div v-else>
+            {{ l }}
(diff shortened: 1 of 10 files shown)


libs/frontend/ui-widgets/src/lib/form/PSMultiselect.vue
```diff
@@ -97,6 +100,10 @@ const props = withDefaults(
     tooltip?: string
     emptyText?: string
     openInitial?: boolean
+    // If true, the label is replaced with a slot
+    customLabel?: boolean
+    // Allow any search input to be selected, even if it's not in the list
+    allowAny?: boolean
```
sampullman, line 106: @abemscac I added this so it's possible to set a Google font which is not included in the list. It seems to work, but let me know if you run into any problems with fonts!

PR #64, opened 2023-11-17
866 cannot drop component at bottom bug

abemscac:
[#866](https://github.com/samatechtw/pubstudio/issues/866)

Hi @sampullman 

Except for the builder D&D "shaking" on hover issue, I also tried to fix the issue where we can't drop a component on root component in this PR.

### DIFF
--- a/libs/frontend/feature-build/src/components/BuildDndOverlay.vue
+++ b/libs/frontend/feature-build/src/components/BuildDndOverlay.vue
@@ -1,0 +1,124 @@
+<template>
+  <div class="build-dnd-overlay" :style="overlayStyle" />
+</template>
+
+<script lang="ts" setup>
+import { computed } from 'vue'
+import { runtimeContext } from '@pubstudio/frontend/util-runtime'
+import { useBuild } from '../lib/use-build'
+
+const hoverOutlineThickness = 3
+const outlineColor = '#F82389'
+
+const { editor } = useBuild()
+
+const overlayStyle = computed(() => {
+  const buildDndState = runtimeContext.buildDndState.value
+  if (buildDndState) {
+    const buildContentWindowInner = document.getElementById('build-content-window-inner')
+    const hoverElement = document.getElementById(buildDndState.hoverCmpId)
+
+    if (buildContentWindowInner && hoverElement) {
+      const {
+        hoverSelf,
+        hoverTop,
+        hoverRight,
+        hoverBottom,
+        hoverLeft,
+        hoverCmpParentIsRow,
+      } = buildDndState
+
+      const { top: buildContentWindowTop, left: buildContentWindowLeft } =
+        buildContentWindowInner.getBoundingClientRect()
+
+      const {
+        top: elementTop,
+        right: elementRight,
+        bottom: elementBottom,
+        left: elementLeft,
+        width: elementWidth,
+        height: elementHeight,
+      } = hoverElement.getBoundingClientRect()
+
+      const builderScale = editor.value?.builderScale ?? 1
+      const outlineThickness = Math.min(
+        hoverOutlineThickness,
+        hoverOutlineThickness * builderScale,
+      )
+
+      const elementRelativeTop = elementTop - buildContentWindowTop
+      const elementRelativeRight = elementRight - buildContentWindowLeft
+      const elementRelativeBottom = elementBottom - buildContentWindowTop
+      const elementRelativeLeft = elementLeft - buildContentWindowLeft
+
+      let [overlayTop, overlayLeft, overlayWidth, overlayHeight] = [0, 0, 0, 0]
+      let [overlayBackgroundColor, overlayBorder] = ['', '']
+
+      if (hoverSelf) {
+        overlayTop = elementRelativeTop
+        overlayLeft = elementRelativeLeft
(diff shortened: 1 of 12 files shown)


libs/frontend/feature-build/src/components/BuildDndOverlay.vue
```diff
@@ -0,0 +1,124 @@
...
+import { computed } from 'vue'
+import { runtimeContext } from '@pubstudio/frontend/util-runtime'
+import { useBuild } from '../lib/use-build'
+
+const hoverOutlineThickness = 3
+const outlineColor = '#F82389'
+
+const { editor } = useBuild()
+
+const overlayStyle = computed(() => {
```
abemscac, line 15: This is the function that calculates the offset&size of the D&D hover overlay.
It seems complicated, maybe there's a simpler way to do this. 😬

libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
```diff
@@ -167,30 +179,57 @@ export const BuilderDndComponent = defineComponent({
...
+        runtimeContext.buildDndState.value = {
+          hoverSelf: dndState.value.hoverSelf,
+          hoverTop: dndState.value.hoverTop,
+          hoverRight: dndState.value.hoverRight,
+          hoverBottom: dndState.value.hoverBottom,
+          hoverLeft: dndState.value.hoverLeft,
+          hoverCmpId: component.value.id,
+          hoverCmpParentIsRow: dndState.value.parentIsRow,
+        }
+      }
```
abemscac, line 225: We only want to show the overlay when a component is hovered in the editor rather than in the component tree, so I add a state to keep track of the D&D state in runtime context, and only update it in `builder-dnd-component`.

sampullman:
https://github.com/samatechtw/pubstudio-builder/assets/550982/e05b327d-7b9f-4218-aade-9cbe7e3c7745

@abemscac I haven't had a chance to take a close look at the code yet, but I can't seem to get drag before/after a component to work, am I missing something?

Drop onto root works, except when the component is already a child of the root. This isn't a big problem, but it would be nice to have.

abemscac:
Oh, sorry about that! 😬 

I think the bug happens because we only check for the value of `flex-direction` to see if a component is row flexbox layout (the `isRowLayout` function).
Since the default style for root component is `{ width: '100%', height: '100%' }`, and it's a `div` at the same time, it should not be considered a row flexbox layout. But in the current implementation `isRowLayout` will return `true` in this case. That means we can only drop a component to the left&right side of the children of root component.

The latest push should've fixed it; sorry about that!

libs/frontend/feature-build/src/components/BuildDndOverlay.vue
sampullman: It makes sense to me!

libs/frontend/feature-build/src/components/BuildDndOverlay.vue
```diff
@@ -0,0 +1,121 @@
...
+
+      const elementRelativeTop = elementTop - buildContentWindowTop
+      const elementRelativeRight = elementRight - buildContentWindowLeft
+      const elementRelativeBottom = elementBottom - buildContentWindowTop
+      const elementRelativeLeft = elementLeft - buildContentWindowLeft
+
+      let [overlayTop, overlayLeft, overlayWidth, overlayHeight] = [0, 0, 0, 0]
+      let [overlayBackgroundColor, overlayBorder] = ['', '']
+
+      overlayTop = elementRelativeTop
```
sampullman, line 58: I factored these out to help de-clutter a little, I don't think it breaks anything but you might want to double check!

libs/frontend/feature-build/src/components/BuildDndOverlay.vue
```diff
@@ -0,0 +1,121 @@
...
+  <div class="build-dnd-overlay" :style="overlayStyle" />
+</template>
+
+<script lang="ts" setup>
+import { computed } from 'vue'
+import { runtimeContext } from '@pubstudio/frontend/util-runtime'
+import { useBuild } from '../lib/use-build'
+
+const hoverOutlineThickness = 6
+const outlineColor = '#000'
```
sampullman, line 11: I experimented a little with the outline colors, so far this worked best for me. I made side-hover borders little thicker, and separated the colors since it was a little confusing. I added a white border style to help indicate when the background is black. It doesn't look perfect, but seems to work.

Eventually we might be able to check the hovered component style and choose a single hover color based on that, but I don't think it's necessary now.

libs/frontend/feature-build/src/components/BuildDndOverlay.vue
abemscac: Hi @sampullman 

I think everything still works after the refactor! One nitpick is that we could probably change this to `let [overlayTop, overlayLeft, overlayWidth, overlayHeight] = [elementRelativeTop, elementRelativeLeft, 0, 0]`, but that doesn't really make anything better, though! 🤔

sampullman: That makes sense! But I think it will cause prettier to expand the statement over 5 lines 😬

PR #67, opened 2023-11-17
fix: css pseudostyle editor bug

### LINKED ISSUE #66: CSS pseudostyle editor bug
When a non-default pseudo style is selected in the editor, all the normal stylings disappear.

https://github.com/samatechtw/pubstudio-builder/assets/550982/96f6843f-861c-4640-81b5-b31e047de303

abemscac:
Close #66

### DIFF
--- a/libs/frontend/ui-widgets/src/lib/InfoBubble.vue
+++ b/libs/frontend/ui-widgets/src/lib/InfoBubble.vue
@@ -131,6 +131,7 @@ const hideBubble = () => {
   @mixin size 20px;
   cursor: pointer;
   position: relative;
+  flex-shrink: 0;
   &:hover {
     :deep(svg g) {
       fill: $color-light1;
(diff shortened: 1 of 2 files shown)


libs/frontend/ui-widgets/src/lib/InfoBubble.vue
```diff
@@ -131,6 +131,7 @@ const hideBubble = () => {
   @mixin size 20px;
   cursor: pointer;
   position: relative;
+  flex-shrink: 0;
```
abemscac, line 134: Hi @sampullman 

I noticed that the inherited from info bubble is invisible when style row is too wide, so I add `flex-shrink: 0` here to make it visible again.

Screenshot 2023-11-17 at 5 55 14 PM

sampullman: Thanks! I think there are probably a few more similar bugs here and there.

PR #79, opened 2023-11-19
77 parent modify children

### LINKED ISSUE #77: Parent component modify children style
Implement a method for parent components to modify/override child styles.

One use for this feature is to modify child style when parent is hovered.

If this is too complicated to implement quickly, improvements to behavior UX can be considered first.

sampullman:
Close #77


libs/frontend/util-component/src/lib/compute-component-breakpoint-styles.ts
```diff
@@ -109,3 +163,29 @@ const computeCustomBreakpointStyles = (
 
   return result
 }
+
+export const computeComponentOverrideStyles = (
```
sampullman, line 167: @abemscac The changes in this PR add a way for a component to modify the style of its children. I was able to reuse a lot of code from computing styles, but there is probably a way to make all of this a little simpler.

There are two problems with my implementation:
1. When a component is deleted or moved, the parent override styles now point to a non-existent component, and are not visible in the UI. I plan to solve this by making sure the styles are still visible, so the user could manually delete them. Another option is to delete the style in add/move commands (and add it back on undo), but it's a little complicated and I'd rather come back to it later.
2. Only specific children can be selected, but it's also useful to select all/some children. I think we can introduce what would basically be real CSS classes to improve this, but I want to avoid that until the rest of the app is done.

abemscac: Got it, I'll take a look at this.
Thanks for letting me know!

PR #84, opened 2023-11-20
46 set custom page title

### LINKED ISSUE #46: Set custom page title
Currently the title of the page defaults to the editor page name.

Allow users to override the title in the theme/meta popup. Adjust the render to check for `page.head.title` and fall back to the page name.

abemscac:
Close #46 
[#900](https://github.com/samatechtw/pubstudio/issues/900)

### DIFF
--- a/apps/web-site/src/app/App.vue
+++ b/apps/web-site/src/app/App.vue
@@ -172,6 +172,9 @@ body {
 p {
   margin: 0;
 }
+.pm-p {
+  min-height: 18px;
+}
 
 a {
   text-decoration: none;
--- a/libs/frontend/feature-build/src/components/StyleRow.vue
+++ b/libs/frontend/feature-build/src/components/StyleRow.vue
@@ -36,15 +36,11 @@
     <div class="label">
       {{ propertyText }}
     </div>
-    <div class="item">
-      <div class="edit-item" :class="{ ['edit-item-inherited']: inheritedFrom }">
-        <div class="value-preview" :class="{ error }">
-          {{ model.value }}
-        </div>
-        <Edit class="edit-icon" @click="edit" />
-      </div>
-      <Minus v-if="!inheritedFrom" class="item-delete" @click.stop="removeStyle" />
+    <div class="value-preview" :class="{ error, ['value-inherited']: inheritedFrom }">
+      {{ model.value }}
     </div>
+    <Edit class="edit-icon" @click="edit" />
+    <Minus v-if="!inheritedFrom" class="item-delete" @click.stop="removeStyle" />
     <InfoBubble v-if="inheritedFrom" class="inherited-from" :message="inheritedFrom" />
   </div>
 </template>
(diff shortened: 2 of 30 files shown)


apps/web-site/src/app/App.vue
```diff
@@ -172,6 +172,9 @@ body {
 p {
   margin: 0;
 }
+.pm-p {
```
sampullman, line 175: Since I changed the prosemirror wrapper element to div, we need this to make line breaks work. There's probably a better way!

libs/frontend/feature-build/src/components/StyleRow.vue
```diff
@@ -36,15 +36,11 @@
...
     </div>
-    <div class="item">
-      <div class="edit-item" :class="{ ['edit-item-inherited']: inheritedFrom }">
-        <div class="value-preview" :class="{ error }">
-          {{ model.value }}
-        </div>
-        <Edit class="edit-icon" @click="edit" />
-      </div>
-      <Minus v-if="!inheritedFrom" class="item-delete" @click.stop="removeStyle" />
+    <div class="value-preview" :class="{ error, ['value-inherited']: inheritedFrom }">
```
sampullman, line 39: I tried to simplify layout and style in a few right menu components, while fixing some UI bugs.

libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
```diff
@@ -53,7 +53,7 @@ const mountProseMirrorEditor = (
     editor.editView?.destroy()
     editor.editView = createComponentEditorView(
       {
-        content: selectedComponent.content ?? '',
+        content: selectedComponent.content || '<div class="pm-p"></div>',
```
sampullman, line 56: I did this to try making empty prosemirror components selectable, but it doesn't work. I'll add an issue later for improving this.

libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
```diff
@@ -100,7 +100,10 @@ watch(gradientColorStyle, () => {
 })
 
 onBeforeUnmount(() => {
-  if (editor.value) {
+  if (editor.value?.editView) {
+    if (editor.value.editView.state.doc.content.size <= 2) {
```
sampullman, line 104: I noticed that if you add some text in a container/footer/etc, then delete it with the prosemirror editor, we can't drag and drop into that component anymore. The reason is that there is an empty `<div></div>` wrapper, which counts as content, so the drop logic fails. This code completely clears the content if the editor is empty.

Another solution could be to allow dropping into components with content, or detecting that the content is "empty."

libs/frontend/feature-site-translations/src/lib/use-site-translations.ts
```diff
@@ -129,7 +129,12 @@ export const useSiteTranslations = (): ISiteTranslationsFeature => {
     if (edit.key !== edit.originalKey) {
       updates[edit.originalKey ?? ''] = undefined
     }
-    setTranslations(getActiveI18n(), updates, true)
+    setTranslations({
+      code: getActiveI18n(),
+      translations: updates,
+      replace: true,
+      forceSave: true,
```
sampullman, line 136: I added this extra parameter since I realized `replace` doesn't trigger a save. So at the end of editing a translation, it wouldn't be saved until the next command was issued.

libs/frontend/util-edit-text/src/lib/schema-basic.ts
```diff
@@ -33,11 +33,11 @@ export const nodes = {
   } as NodeSpec,
 
   /// A plain paragraph textblock. Represented in the DOM
-  /// as a `<p>` element.
+  /// as a `<div>` element so that i18n nodes can be nested
```
sampullman, line 36: I changed the wrapper component for text from `p` to `div`. The main reason is that since `i18n` entries are also edited with ProseMirror, including a translation in a component  would result in nested `<p>`s, which is not allowed.

I'm still not that happy with how I implemented the `i18n` feature, but it's getting better.

PR #92, opened 2023-11-23
fix: image hover after resize bug

### LINKED ISSUE #69: Image hover after resize bug
After adjusting the width and height percentages of a photo, the hovering over the object causes erratic jumping.

https://github.com/samatechtw/pubstudio-builder/assets/550982/0f03175f-c9e1-4b3e-857e-153bbb8836f0

abemscac:
Close #69

### DIFF
--- a/libs/frontend/feature-build-event/src/lib/use-build-event.ts
+++ b/libs/frontend/feature-build-event/src/lib/use-build-event.ts
@@ -33,6 +33,7 @@ import {
   calcNextWidth,
   getHeightPxPerPercent,
   getWidthPxPerPercent,
+  isLengthValue,
 } from './util-resize'
 
 const clickEventType = document.ontouchstart !== null ? 'click' : 'touchend'
@@ -220,7 +221,8 @@ export const useBuildEvent = () => {
       // This is for elements that use an additional wrapper for hover edges, such as img.
       // Kebab case "component-id" will be converted to camel case "componentId".
       // See https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
-      resolveComponent(site.value.context, target?.dataset.componentId)
+      resolveComponent(site.value.context, target?.dataset.componentId) ??
+      editor?.resizeData?.component
     if (editor) {
       const data = editor.resizeData
       if (component && data) {
@@ -342,7 +344,12 @@ export const useBuildEvent = () => {
       const resizeCmpRect = resizeCmp.getBoundingClientRect()
       const parentRect = (resizeCmp.parentElement as HTMLElement).getBoundingClientRect()
 
-      const startHeight = initialHeightProp || `${resizeCmpRect.height}px`
+      let startHeight = initialHeightProp || `${resizeCmpRect.height}px`
+      if (!isLengthValue(startHeight)) {
+        // Use the current height (px) of the resized component as the inital height
+        startHeight = `${resizeCmpRect.height}px`
+      }
+
       let startWidth = initialWidthProp
       if (!startWidth) {
         const resizeCmpIsAbsolute =
@@ -354,6 +361,9 @@ export const useBuildEvent = () => {
         } else {
           startWidth = '100%'
         }
+      } else if (!isLengthValue(startWidth)) {
+        // Use the current width (px) of the resized component as the inital width
+        startWidth = `${resizeCmpRect.width}px`
       }
 
       const heightUnit = startHeight.replace(/[-\d.]/g, '') as CssUnit
--- a/libs/frontend/feature-build-event/src/lib/use-build-event.ts
+++ b/libs/frontend/feature-build-event/src/lib/use-build-event.ts
@@ -346,7 +346,7 @@ export const useBuildEvent = () => {
 
       let startHeight = initialHeightProp || `${resizeCmpRect.height}px`
       if (!isLengthValue(startHeight)) {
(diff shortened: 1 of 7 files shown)


libs/frontend/feature-build-event/src/lib/use-build-event.ts
```diff
@@ -342,7 +344,12 @@ export const useBuildEvent = () => {
       const resizeCmpRect = resizeCmp.getBoundingClientRect()
       const parentRect = (resizeCmp.parentElement as HTMLElement).getBoundingClientRect()
 
-      const startHeight = initialHeightProp || `${resizeCmpRect.height}px`
+      let startHeight = initialHeightProp || `${resizeCmpRect.height}px`
+      if (!isLengthValue(startHeight)) {
+        // Use the current height (px) of the resized component as the inital height
+        startHeight = `${resizeCmpRect.height}px`
+      }
```
abemscac, line 351: This is to fix a bug when the resize happens when the width/height of a component is not a length value (i.e. `auto`, `initial`, `inherit`, etc); that'll cause the width/height to be something like `0auto` or `NaNinitial` during the process.

libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
```diff
@@ -86,7 +77,12 @@ const renderImageHover = (
   // Pass the margin of img to the wrapper element, and set the margin of the rendered img
   // to 0 during hover to avoid shifting issue.
   const imgStyleProp = structuredClone(props.style ?? {}) as Record<string, unknown>
-  imgStyleProp['margin'] = '0'
+  // Set the width&height of <img> to 100% because the width&height of the hover wrap (parent) is defined.
+  // In HTML, if the width or height of the parent of <img> is a length value (i.e. 100%, 100px, 1rem, etc.),
+  // the % in the dimensions of <img> will be calculated based on the size of its parent element.
+  imgStyleProp.width = '100%'
+  imgStyleProp.height = '100%'
+  imgStyleProp.margin = '0'
```
abemscac, line 85: The image hover issue is taking way longer than I expected 😵. Sorry about that!
The best solution I could find is to get the width&height directly from the bounding rect of the target ``, and use it as the dimensions of the hover wrap. By setting the width&height of the `` to 100%, it'll have exactly the same width&height of its parent element due to how `` works in HTML. I didn't find any information about this on MDN, though. 🤔 

I always thought the width&height (%) of an absolute positioning `` is calculated by the dimensions of its `position: relative` parent, which is actually wrong. No matter if the `` is absolute positioned or not, if the width/height of its direct parent is not a length value, `width: 10%` on a `` will result in `sourceWidth * 10%`. But as long as the width/height of its parent is a length value (i.e. `100px`, `50%`, `0.5rem`, etc), `width: 10%` on a `` will then result in `parentWidthInPx * 10%`.

libs/frontend/feature-build-event/src/lib/use-build-event.ts
```diff
@@ -220,7 +221,8 @@ export const useBuildEvent = () => {
       // This is for elements that use an additional wrapper for hover edges, such as img.
       // Kebab case "component-id" will be converted to camel case "componentId".
       // See https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
-      resolveComponent(site.value.context, target?.dataset.componentId)
+      resolveComponent(site.value.context, target?.dataset.componentId) ??
+      editor?.resizeData?.component
```
abemscac, line 225: This is to fix a render issue when the cursor is not hovering on the image during resize.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/a2e922c2-76f3-4bd0-9083-f840c0173912

libs/frontend/feature-build-event/src/lib/use-build-event.ts
sampullman: 👍

libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
sampullman: Yeah! I think I have encountered this problem in the past. CSS interactions with auto/initial dimensions can be unexpected, especially when combined with flex containers.

There is still a small jump in height when selecting an image and hovering away, but overall this seems like a big improvement, I think we can move forward!


https://github.com/samatechtw/pubstudio-builder/assets/550982/3ddf88a7-6032-41c8-8095-5ac1d755f986

sampullman:
Thanks! I noticed the `main` branch CI was failing in the last PR, so I included a patch here.

libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
abemscac: Oh, sorry about that! I think the issue is caused by the `Math.floor()` at `builder-dnd-component:57` and `builder-dnd-component:58` when calculating the width&height to be used on the wrapper div. Removing them seems to make the small jump not that obvious:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/eb4517c2-d4bf-4478-a779-b26b186b64a4

I'll include this in my next PR, or you could do that if that'll make things easier for you, thank you!

sampullman: You can include it, no problem! I am stuck working on my own bug at the moment 😅

PR #93, opened 2023-11-23
feat: select asset as src of image component

### LINKED ISSUE #43: Select asset as src of image component
Currently, the only way to add an asset in the editor is to drag/select in the asset menu, or manually copy-paste the URL to an image src.

It should be possible to select an asset from the image `src` property in the right menu. Add a button that shows a modal with all the site assets. Filter by the current site by default, but allow the user to set the filter (similar to asset upload). When an asset is clicked, close the modal and set the image `src` to the asset URL.

abemscac:
Close #43

### DIFF
--- a/libs/frontend/feature-build/src/components/component-menu/ComponentInputRow.vue
+++ b/libs/frontend/feature-build/src/components/component-menu/ComponentInputRow.vue
@@ -21,6 +21,7 @@
         {{ value }}
       </div>
       <Edit class="edit-icon" @click="edit" />
+      <slot />
       <Settings v-if="showEditInput" class="edit-icon" @click="emit('editInput')" />
     </template>
   </div>
@@ -106,6 +107,7 @@ const datalist = computed<IDatalistOption[] | undefined>(() => {
 .value-preview {
   @mixin truncate;
   @mixin text 14px;
+  flex-grow: 1;
 }
 
 .edit-icon {
--- a/libs/frontend/feature-build/src/components/component-menu/ComponentInputRow.vue
+++ b/libs/frontend/feature-build/src/components/component-menu/ComponentInputRow.vue
@@ -104,6 +104,10 @@ const datalist = computed<IDatalistOption[] | undefined>(() => {
 <style lang="postcss" scoped>
 @import '@theme/css/mixins.postcss';
 
+.label {
+  @mixin text 14px;
+}
+
 .value-preview {
   @mixin truncate;
   @mixin text 14px;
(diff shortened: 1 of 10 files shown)


libs/frontend/feature-build/src/components/component-menu/ComponentInputRow.vue
```diff
@@ -106,6 +107,7 @@ const datalist = computed<IDatalistOption[] | undefined>(() => {
 .value-preview {
   @mixin truncate;
   @mixin text 14px;
+  flex-grow: 1;
```
abemscac, line 110: This is to fix an issue when input value is so short that buttons are not pushed to the end/right.

Screenshot 2023-11-23 at 3 57 29 PM

libs/frontend/feature-site-assets/src/components/SelectAssetModal.vue
```diff
@@ -0,0 +1,288 @@
...
+)
+const { show, initialSiteId } = toRefs(props)
+
+const emit = defineEmits<{
+  (e: 'cancel'): void
+  (e: 'select', asset: ISiteAssetViewModel): void
+}>()
+
+const rootApi = inject(ApiInjectionKey) as PSApi
+const api = useSiteAssetApi(rootApi)
```
abemscac, line 118: Hi @sampullman 

I didn't reuse the `useSiteAssets` composable here because asset menu is using the same composable as well. Using it to fetch site assets will cause the existing result in asset menu to be erased.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/f8aed7c8-54de-4411-8b8b-a146bcbff49d

libs/frontend/feature-site-assets/src/components/SelectAssetModal.vue
```diff
@@ -0,0 +1,288 @@
...
+  }
+}
+
+const listAssets = async () => {
+  loading.value = true
+  try {
+    const response = await api.listSiteAssets({
+      ...filter.value,
+      user_id: store.user.id.value,
+      state: SiteAssetState.Uploaded,
```
abemscac, line 182: The modal won't show any assets in development mode because there's no site assets in `Uploaded` state in the fixture. You can comment out this line (182) to temporarily see what does it look like when there's asset in the modal.

Screenshot 2023-11-23 at 4 21 42 PM

libs/frontend/feature-build/src/components/component-menu/ComponentInputRow.vue
sampullman: Thanks! I've been trying to simplify the layout/css here and there, and I missed that case.

Issue #80, opened 2023-11-20
Show deleted/moved child styles


sampullman:
When a component is deleted or moved, the parent override styles now point to a non-existent component, and are not visible in the UI. 

This can be solvedby making sure the styles are still visible in the `Child Styles` dropdown, so the user could manually delete them. In the dropdown, we can highlight the options red if they correspond to missing children.

### Notes

Another option is to delete the style in add/move commands (and add it back on undo), but it's a little complicated and I'd rather come back to it later.

PR #96, opened 2023-11-24
fix: prosemirror component empty bug

### LINKED ISSUE #85: ProseMirror component empty bug
When all the text is removed from a prosemirror/text component, it cannot be edited again unless text manually is manually added in the right component menu.

In general, it is hard to select small components, but this is more critical since there's no other way to start editing text.

We should also try to show the cursor when it's at the edge (beginning or end) of the prosemirror editor (padding?).

https://github.com/samatechtw/pubstudio-builder/assets/550982/b817ab0a-5f60-480f-8fa9-40e16935af54

abemscac:
Close #85

### DIFF
--- a/apps/web-site/src/app/App.vue
+++ b/apps/web-site/src/app/App.vue
@@ -169,11 +169,12 @@ body {
   }
 }
 
+.pm-p:after {
+  content: '\200b';
+}
+
 p {
   margin: 0;
-}
-.pm-p {
-  min-height: 18px;
 }
 
 a {
--- a/libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
+++ b/libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
@@ -24,7 +24,7 @@ const props = defineProps<{
 }>()
 
 const { component, editor } = toRefs(props)
-const { site, editComponent } = useBuild()
+const { site } = useBuild()
 
 const container = ref<HTMLDivElement>()
 const containerId = computed(() => getProseMirrorContainerId(component.value))
@@ -104,13 +104,8 @@ watch(gradientColorStyle, () => {
 })
 
 onBeforeUnmount(() => {
-  if (editor.value?.editView) {
-    if (editor.value.editView.state.doc.content.size <= 2) {
-      editComponent(component.value, { content: '' })
-    }
-    editor.value.editView?.destroy()
-    editor.value.editView = undefined
-  }
+  editor.value.editView?.destroy()
+  editor.value.editView = undefined
 })
 
 watch(
--- a/libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
+++ b/libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
@@ -54,8 +54,8 @@ const renderImageHover = (
     imgElement?.getBoundingClientRect() ?? {}
 
   const { builderScale = 1 } = site.editor ?? {}
-  const unscaledWidth = Math.floor(rectWidth / builderScale)
-  const unscaledHeight = Math.floor(rectHeight / builderScale)
+  const unscaledWidth = rectWidth / builderScale
+  const unscaledHeight = rectHeight / builderScale
 
   width = `${unscaledWidth}px`
   height = `${unscaledHeight}px`
@@ -94,7 +94,9 @@ const renderImageHover = (
     {
       'data-component-id': component.id,
       style: {
-        display: 'flex',
+        // Use inline-block to simulate the bottom space caused by the inline behavior of <img>.
+        // See https://stackoverflow.com/a/5804278/19772349 for more information.
+        display: 'inline-block',
         width,
         height,
         margin,
--- a/libs/frontend/feature-build/src/components/StyleMenuEdit.vue
+++ b/libs/frontend/feature-build/src/components/StyleMenuEdit.vue
@@ -81,7 +81,6 @@ const editing = (propName: string) => {
 }
 
 const escPress = () => {
-  console.log('ESCCC')
   editingMenu.value = false
 }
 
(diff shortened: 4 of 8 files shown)


apps/web-site/src/app/App.vue
```diff
@@ -172,9 +172,6 @@ body {
 p {
   margin: 0;
 }
-.pm-p {
-  min-height: 18px;
-}
```
abemscac, line 177: Hi @sampullman 

I'm not sure if I'm getting this wrong; I think the prosemirror editor is only visible in the builder, but it seems that currently the styles of `.pm-p` is defined in `apps/web-site/src/app/App.vue` and `pubstudio-builder/libs/frontend/feature-build/src/components/preview/SitePreview.vue`.

I moved the style definition of `.pm-p` to `apps/web/src/app/App.vue` instead so that only the builder app is affected. A new [PR](https://github.com/samatechtw/pubstudio/pull/927) is created in the pubstudio repo for this purpose.

libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
```diff
@@ -104,13 +104,8 @@ watch(gradientColorStyle, () => {
...
 onBeforeUnmount(() => {
-  if (editor.value?.editView) {
-    if (editor.value.editView.state.doc.content.size <= 2) {
-      editComponent(component.value, { content: '' })
-    }
-    editor.value.editView?.destroy()
-    editor.value.editView = undefined
-  }
+  editor.value.editView?.destroy()
+  editor.value.editView = undefined
```
abemscac, line 108: I noticed that the [fix](https://github.com/samatechtw/pubstudio-builder/pull/84#discussion_r1400702449) for the prosemirror content cleared issue would make the component keep getting focused on unmount due to the "auto focus on affected component" feature in all commands. We're unable to select another component when we clear out the text in a prosemirror editor.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/a455e5bc-bc54-4ae1-9235-9d97e6510f8b

After trying out for a while I think it's easier to just use update the content of the target component to `undefined` in the transaction of a prosemirror editor. I adjusted `editorStateToHtml` to make it work.

libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
```diff
@@ -94,7 +94,9 @@ const renderImageHover = (
     {
       'data-component-id': component.id,
       style: {
-        display: 'flex',
+        // Use inline-block to simulate the bottom space caused by the inline behavior of <img>.
+        // See https://stackoverflow.com/a/5804278/19772349 for more information.
+        display: 'inline-block',
```
abemscac, line 99: I noticed that there's still a strange jump issue even though the `Math.floor` when calculating the dimensions of the image hover wrap has been removed.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/c40a9e57-555e-4c8b-977c-ed0d89721e1f

After some investigation I learned that the `` in HTML is in inline display, so it actually "preserves" some spaces at the bottom for some characters. After changing the `display` of the hover wrap from `flex` to `inline-block`, the problem seems to be solved!

abemscac:
Hi @sampullman 

This is not related to this issue, but I noticed that there'll be an error about "MutationObserver" (I've never heard that before!) by doing the following actions:

1. Add whatever component that can have text as content (e.g. Text).
2. Delete the component.
3. Press undo to recover the component.
4. Clear out the text in the component.
5. Click at somewhere else to unfocus the component; the error is now logged in the console.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/424dbee0-ccf1-483f-84d6-2913766994ab

It may seem that the problem is caused by prosemirror editor, but I found that comment out/remove `pubstudio-builder/libs/frontend/feature-build/src/components/component-menu/ComponentTabStyle.vue:4` (which is `<ComponentChildStyles v-if="!!component.children?.length" :component="component" />`) actually makes the error gone. I don't understand why, I'll try to solve it when doing #80, if it's not solved by that time!

apps/web-site/src/app/App.vue
sampullman: This should be left in, otherwise a double line break will not show the correct spacing in the live version. `.pm-p` is the class added to each div generated by ProseMirror. You are correct that we need it in the `pubstudio` repo as well, though.

However, I realized an obvious mistake in my implementation - it doesn't account for different line heights! Below is the live version (without `pm-p` style), editor version (with `pm-p` style), and the editor version with selected component.

space

I think I've found a solution though, we can use a zero width character to the editor/live style:
```css
.pm-p:after {
  content: '\200b';
}
```

This will cause the selected/active ProseMirror editor to have line breaks that are double height, but it can be solved by setting the ProseMirror editor line break to `display: none;`. Hopefully I didn't miss anything 🤞

libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
sampullman: Makes sense, I probably should have thought of that in the first place!

libs/frontend/feature-build/src/lib/dnd/builder-dnd-component.ts
sampullman: Good catch!

sampullman:
Nice work, the editor has improved a lot this week!

sampullman:
> It may seem that the problem is caused by prosemirror editor, but I found that comment out/remove pubstudio-builder/libs/frontend/feature-build/src/components/component-menu/ComponentTabStyle.vue:4 (which is <ComponentChildStyles v-if="!!component.children?.length" :component="component" />) actually makes the error gone. I don't understand why, I'll try to solve it when doing https://github.com/samatechtw/pubstudio-builder/issues/80, if it's not solved by that time!

Interesting, I probably did something terrible when implementing `ComponentChildStyles` 🤔 

Actually that reminds me, I keep forgetting to create an issue for something else I saw in the debug tools:

Screenshot 2023-11-17 at 3 05 49 AM

It seems like every component has this, maybe we are abusing Vue is some way. It might be worth solving as well, since it's probably not great for performance to have all these errors/warnings on every component.

PR #98, opened 2023-11-25
feat: component tree nesting and fix hidden #71

### LINKED ISSUE #71: Component tree deep nesting
When there are deeply nested components, they get pushed to the right in the component tree and cannot be viewed.

Come up with a way to make sure all components can be seen in the tree:
- Horizontal scrolling could work, but it's a little awkward
- Increasing the width of the tree works, but it's inconvenient for the editor

Screenshot 2023-10-14 at 3 17 46 PM

sampullman:
Close #71


libs/frontend/feature-build/src/components/ComponentTree.vue
```diff
@@ -1,35 +1,167 @@
...
+      }
+      if (node.cmp.children) {
+        stack.push(...node.cmp.children.map((c) => makeNode(c, depth + 1)))
+      }
+    }
+  }
+  return depth
+})
+
+const treeStyles = computed(() => {
```
sampullman, line 64: @abemscac I "fixed" the nesting problem by checking the deepest component level, and setting some styles based on that. It's kind of complicated and requires traversing most of the component tree to calculate it, but seems to work.

I also made some changes to (hopefully) improve performance. I moved all the CSS out of the `ComponentTreeItem`, so it only needs to be processed once. Since we know the max tree depth, it's also possible to dynamically create styles for all the title paddings at each level.

While implementing this, I also re-enabled the eye/hide feature, although I'm not sure I like it.

PR #111, opened 2023-11-27
fix: show deleted/moved child styles

### LINKED ISSUE #80: Show deleted/moved child styles

abemscac:
Close #80 

Hi @sampullman 

I hope this works; eventually I still made remove component override style an individual command so that we can get all the style entries back when pressing undo.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/12d8d0c4-abdf-4a66-bf2f-63209dba5be0

One (of the many 😬) thing that can still be improved improve in my implementation is, I tried to re-select the deleted selector from the dropdown when a removal has been redo (similar to how we focus on the affected component in a command), but I found the code being very ugly because I have to use `watch` to observe the changes of the selected component id and `component.value.style.overrides`. I ended up giving up on that and wonder if you would have any good idea, sorry about that 😵!


libs/frontend/feature-build/src/components/component-menu/ComponentChildStyles.vue
```diff
@@ -37,20 +61,21 @@
         :key="`${entry.pseudoClass}-${entry.property}-${entry.value}`"
         :style="entry"
         :editing="editing(entry.property)"
+        :inheritedFrom="getInheritedFrom(entry)"
```
abemscac, line 64: Add the inherited from here so that it shows on overridden styles as well. The style inheritance in overridden styles is working already; it's just that the UI was not showing where is the style entry inherited from.

libs/frontend/feature-build/src/components/component-menu/ComponentChildStyles.vue
```diff
@@ -86,14 +112,39 @@ const props = defineProps<{
...
+  const overriddenChildrenIds = Array.from(overriddenChildrenIdSet.value)
+  return overriddenChildrenIds.some((id) => !childrenIdSet.value.has(id))
 })
 
+const currentSelectorMissing = computed(
+  () =>
+    !!selector.value &&
+    !component.value.children?.some((child) => child.id === selector.value),
+)
+
```
abemscac, line 147: The code here seems kind of messy 😬.
My intent was to:

- Be able mark invalid selectors in the menu (selectors where the corresponding children are not there anymore).
- Be able to show the remove button when the selected selector is invalid.
- Use `set.has` instead of `array.includes` to avoid iterating arrays too much, but I'm not sure if it's worth doing it here.

libs/frontend/util-test-mock/src/lib/command/mock-set-component-override-style-data.ts
```diff
@@ -33,7 +33,7 @@ export const mockAddComponentOverrideStyleData = (
   return data
 }
 
-export const mockRemoveComponentOverrideStyleData = (
+export const mockRemoveComponentOverrideStyleEntryData = (
```
abemscac, line 36: I renamed this so that it's not conflicting with the new command.

libs/frontend/feature-build/src/components/component-menu/ComponentChildStyles.vue
sampullman: It's fine, though I don't think it's makes a big difference here. My one note is that it's probably better to avoid going from Array -> Set -> Array, for example `Object.keys(component.value.style.overrides ?? {})` could be calculated once and used in both `overriddenChildrenIdSet` and `hasMissingSelector`.

libs/frontend/feature-build/src/components/component-menu/ComponentChildStyles.vue
sampullman: I might be missing something, but is it actually possible to have overridden child styles? Currently we can't assign mixins to children, though in theory it could be implemented.

sampullman:
This is great, thanks! Everything seems to be working as expected.

libs/frontend/feature-build/src/components/component-menu/ComponentChildStyles.vue
abemscac: I think so! It's just that styles from larger breakpoints would be the only source of inherited styles for child styles.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/81e07697-1e9f-4e61-9d7f-0923b9ee50f3

sampullman: Oh, that's right!

PR #116, opened 2023-11-30
feat: hide new component menu when dragging

### LINKED ISSUE #106: Updating prosemirror text doesn't save
Updating some prosemirror text and clicking away doesn't save the text. It only saves when another action is done.

The cause seems to be the editor save timer overriding the site/page save timer, which causes only the editor to be saved.

https://github.com/samatechtw/pubstudio-builder/assets/550982/4bd691d3-38ed-47a9-a490-4e63e338425c

### LINKED ISSUE #108: Hide the new component menu when dragging
In order to support dropping a new component into the component tree, hide the build->new menu when the user starts to drag a component.

- Show menu again if drag is cancelled (component not dropped to tree/editor)
  - Consider showing the menu again no matter what (e.g. if component is added)
- Cancel drag if escape is pressed (?)

abemscac:
Close #106
Close #108

### DIFF
--- a/libs/frontend/feature-build/src/components/build-menu/BuildMenu.vue
+++ b/libs/frontend/feature-build/src/components/build-menu/BuildMenu.vue
@@ -72,7 +72,7 @@
     <Transition name="submenu">
       <component
         :is="submenu"
-        v-if="!!editor?.buildSubmenu"
+        v-if="showSubmenu"
         class="submenu"
         @openPageMenu="emit('openPageMenu')"
         @showCreate="emit('showCreateModal')"
@@ -99,6 +99,7 @@ import BuildMenuAsset from './BuildMenuAsset.vue'
 import { useBuild } from '../../lib/use-build'
 import BuildMenuBehavior from './BuildMenuBehavior.vue'
 import BuildMenuHistory from './BuildMenuHistory.vue'
+import { dragSource } from '../../lib/dnd/use-drag-drop'
 
 const { t } = useI18n()
 const { editor } = useBuild()
@@ -132,6 +133,17 @@ const toggleSubMenu = (newSubmenu: BuildSubmenu) => {
   const submenu = editor.value?.buildSubmenu
   setBuildSubmenu(editor.value, submenu === newSubmenu ? undefined : newSubmenu)
 }
+
+const showSubmenu = computed(() => {
+  const { buildSubmenu } = editor.value ?? {}
+  if (buildSubmenu === BuildSubmenu.New) {
+    // Only show the new component menu when no new
+    // component is being dragged.
+    return !dragSource.value?.addData
+  } else {
+    return !!buildSubmenu
+  }
+})
 </script>
 
 <style lang="postcss" scoped>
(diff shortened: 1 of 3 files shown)


libs/frontend/feature-build/src/lib/dnd/use-drag-drop.ts
```diff
@@ -281,6 +283,12 @@ export const useDragDrop = (props: IUseDragDropProps): IUseDragDrop => {
         }
       }
     }
+
+    if (site.editor?.buildSubmenu === BuildSubmenu.New && dragSource.value?.addData) {
+      // Hide new component menu when a new component is dropped
+      setBuildSubmenu(site.editor, undefined)
+    }
```
abemscac, line 290: I eventually hide the new component menu when a new component is being dropped in the editor (either in the component tree or in another component). I think this behavior feels more natural to me, but I'm not sure if this also makes sense to you 😬.

abemscac:
Hi @sampullman 

I noticed that typing won't work in the code editor when the editor is empty unless we press "ctrl + a" to select all content beforehand (even though it's empty!). The keydown event did gets fired, but for some reason the state of ProseMirror editor just don't gets updated when a single key is pressed. That being said, the keys we defined in our own keymap (tab, copy, paste, etc) did work without any issue.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/e04db1ec-b9f3-4a40-b9b7-57272762c120

I rolled back to the branch at 20th, Nov (`2e83e37` for pubstudio and `fc98a53` for pubstudio-builder) and found that the input did work back then. I tried to revert everything I could find but the bug still persists 😵. I think the problem lies in somewhere else, but I didn't manage to figure it out. It would be great if you could take a look at this, thank you so much!

libs/frontend/feature-build/src/components/build-render/ProseMirrorEditor.vue
```diff
@@ -160,6 +162,7 @@ watch(gradientColorStyle, () => {
 })
 
 onBeforeUnmount(() => {
+  siteStore.value.save(site.value)
```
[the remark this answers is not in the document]
sampullman: Oh, that makes sense!

Let's go with your solution here, I've also considered saving in `replaceLastCommand` but wasn't sure (for the same reason).

Maybe if it ends up being troublesome we can add a longer timer (like 15/30 seconds) so it eventually gets saved, but doesn't trigger too often.

libs/frontend/feature-build/src/lib/dnd/use-drag-drop.ts
sampullman: I'm fine either way, but I think you're right, it feels a little more natural this way. It could be annoying if you're trying to add a lot of components quickly, but maybe we can add component hotkeys for "power users" later on.

libs/frontend/feature-build/src/components/build-menu/BuildMenu.vue
```diff
@@ -132,6 +133,17 @@ const toggleSubMenu = (newSubmenu: BuildSubmenu) => {
   const submenu = editor.value?.buildSubmenu
   setBuildSubmenu(editor.value, submenu === newSubmenu ? undefined : newSubmenu)
 }
+
+const showSubmenu = computed(() => {
+  const { buildSubmenu } = editor.value ?? {}
+  if (buildSubmenu === BuildSubmenu.New) {
```
sampullman, line 139: I made a slight change to this logic, and hopefully didn't break it!

PR #121, opened 2023-12-01
feat: move margin padding menu to right menu

### LINKED ISSUE #107: Move style menu margin/padding to right menu
Move the style menu margin/padding popup to the right menu, since it overlaps the editor window.

Another option is to make the window "float", so it can be dragged anywhere on the screen. In this case we should store the window location in the editor context.

abemscac:
Close #107

### DIFF
--- a/libs/frontend/util-site-deserialize/src/lib/deserialize-site.ts
+++ b/libs/frontend/util-site-deserialize/src/lib/deserialize-site.ts
@@ -43,6 +43,8 @@ export const deserializeEditor = (
         builderScale: serializedEditor.builderScale,
         cssPseudoClass: serializedEditor.cssPseudoClass,
         templatesShown: serializedEditor.templatesShown,
+        // Added 231201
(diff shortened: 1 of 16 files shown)


abemscac:
Hi @sampullman 

I think this PR is ready for review, thank you!

sampullman:
Looks good to me, thanks!

libs/frontend/util-site-deserialize/src/lib/deserialize-site.ts
```diff
@@ -43,6 +43,8 @@ export const deserializeEditor = (
         builderScale: serializedEditor.builderScale,
         cssPseudoClass: serializedEditor.cssPseudoClass,
         templatesShown: serializedEditor.templatesShown,
+        // Added 231201
+        componentMenuCollapses: serializedEditor.componentMenuCollapses ?? {},
```
sampullman, line 47: Instead of making new context fields optional, I've been adding a default in the deserialization function. It's not a big difference, but should make the application code slightly cleaner, and we can include all these "defaults" in a central location.

PR #128, opened 2023-12-04
feat: press enter to rename component in tree

### LINKED ISSUE #42: Press enter to rename component in tree
When focusing on a component-tree-item, pressing Enter changes the text into an input for faster renaming.

Pressing Enter again, or clicking away (tree-item loses focus) causes the name to be saved.

abemscac:
Close #42

### DIFF
--- a/libs/frontend/feature-build-event/src/lib/use-build-event.ts
+++ b/libs/frontend/feature-build-event/src/lib/use-build-event.ts
@@ -34,14 +34,37 @@ import {
   getWidthPxPerPercent,
   isLengthValue,
 } from './util-resize'
+import { runtimeContext } from '@pubstudio/frontend/util-runtime'
 
 const clickEventType = document.ontouchstart !== null ? 'click' : 'touchend'
 
 const isRenderer = (target: HTMLElement | undefined): boolean => {
   return !!target?.closest('.build-content')
 }
-const isComponentTree = (target: HTMLElement | undefined): boolean => {
-  return !!target?.closest('.component-tree')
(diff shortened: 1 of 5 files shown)


libs/frontend/feature-build-event/src/lib/use-build-event.ts
```diff
@@ -136,11 +159,21 @@ export const useBuildEvent = () => {
   const clickRightMenu = (_target: HTMLElement | undefined) => {
     setBuildSubmenu(editor.value, undefined)
   }
-  const clickComponentTree = (_target: HTMLElement | undefined) => {
+  const clickComponentTree = (componentTreeItemId: string) => {
     setBuildSubmenu(editor.value, undefined)
+    runtimeContext.componentTreeItemRenameData.value.treeItemId = componentTreeItemId
```
abemscac, line 164: Hi @sampullman 

I left some comments here and there in this commit to explain what I did; it looks kind of complicated now that I look at it 😬.

I was trying to make the rename experience as close to VSCode as possible (because I think the experience is great!), and I found myself needing some states like "am I focusing on the component tree?" so that we know whether pressing enter should show the rename input or not. For example, if I click on a component from the editor, even though the corresponding tree item is active in the component tree, showing the rename input when enter is pressed in that case would be a little strange because I'm actually looking at the editor now instead of the component tree.

I didn't implement that feature in this commit for I though that may take a while (is it 🤔?). I use some simple state in the runtime context to achieve something similar (check if the last clicked item is a component tree item).

PR #129, opened 2023-12-04
fix: update default site and refactor for script use #122

### LINKED ISSUE #122: Fix default site
The default site should be updated whenever the data format changes, or if new theme variables are added. It might be convenient to add a script to export the default site.

sampullman:
Close #122


tools/scripts/generate-default-site.ts
```diff
@@ -0,0 +1,76 @@
...
+  pseudoClass: CssPseudoClass.Default,
+  property: Css.Margin,
+  value: '40px 0 40px 0',
+})
+
+const printSiteField = (key: string, field: string | null | undefined) => {
+  console.log(`pub const ${key}: &str = r##"${field ?? ''}"##;`)
+}
+
+const stored = storeSite(site.value)
```
sampullman, line 72: @abemscac This was a big refactor for what should have been a simple task, but I think it needed to be done. My goal was to be able to use `useBuild` from regular node scripts, so that it's possible to create and export a site. The problem was that some Vue files were in the dependency tree, as well as the local store. I was able to shift things around to make it work, the main changes are:
- Split components in `feature-build` to `libs/frontend/ui-build`
- Move components rendered in the builder (prosemirror, link tooltip, etc) to `libs/frontend/feature-render-builder`. They can't go in `ui-build` because there would be a circular dependency.
- Move site initialization (mainly `use-local-store` and `use-api-store` to a `libs/frontend/feature-site-store-init`. Now `initializeSite` takes an `ISiteStore` param, so it's possible to pass a mock to avoid local/API stores if you're writing a script (see the beginning of this file). 

If writing these site scripts becomes more common, we can add some initialization/convenience methods to improve the experience.

PR #130, opened 2023-12-05
feat: tab and shift tab to change selection

### LINKED ISSUE #88: Tab and shift-tab to change selection
When a component is selected in the editor and the tab key is pressed, select the next component. The logic is:

1. If the component has 1 or more children, select the first one
2. If the component's parent has a "next" child, select it
3. Otherwise apply the 2. logic to the parent
4. Last component on the page - do nothing

When shift-tab is pressed, select the previous component (reverse of above logic).

abemscac:
Close #88

### DIFF
--- a/libs/frontend/feature-build-event/src/lib/select-component.ts
+++ b/libs/frontend/feature-build-event/src/lib/select-component.ts
@@ -1,0 +1,83 @@
+import {
+  getComponentTreeItemId,
+  setSelectedComponent,
+} from '@pubstudio/frontend/feature-editor'
+import { IComponent, ISite } from '@pubstudio/shared/type-site'
+
+export const selectComponent = (site: ISite, component: IComponent) => {
+  setSelectedComponent(site, component)
+
+  // Scroll to the corresponding tree item if component tree is visible
+  if (site.editor?.showComponentTree) {
+    const treeItemId = getComponentTreeItemId(component)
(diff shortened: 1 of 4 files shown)


abemscac:
Hi @sampullman 

I noticed that the browser would sometimes focus on other elements when pressing tab (i.e. inputs, browser URL input, and buttons in the console if it's opened). `e.stopPropogation()` and `e.preventDefault()` doesn't seem to be able to stop this from happening as well. Maybe this is the default behavior of browsers?

sampullman:
It might be only Chrome that does this, but the reason is that the default tab behavior occurs on keydown, and we're listening for keyup.

I included a temporary solution `;(document.activeElement as HTMLElement)?.blur()`, which seems to work well enough. When tab is pressed in inputs outside the editor, the expected behavior occurs. One other thing I noticed is that we might want to override the tab behavior in the prosemirror editor, but it's not too critical.

I moved the component selection functionality to a separate file. I'm considering moving away from some of the "gobal" usables like `useBuild` and `useBuildEvent`, because I don't think there's much benefit. I'm not completely sure, but I think it uses a lot of memory since copies of the usable's functions are made every time `useBuild` is called.

abemscac:
> It might be only Chrome that does this, but the reason is that the default tab behavior occurs on keydown, and we're listening for keyup.

Ah, I see! How come I didn't think of that.
Thanks for fixing the focus issue!

PR #133, opened 2023-12-07
feat: copy root component

### LINKED ISSUE #44: Copy root component
Currently it is not possible to copy/paste the root component or its styles. Allow users to copy the root component and its styles, and make it possible to replace the root component of a page by pasting a component.

When the root component is selected and a component is copied to the clipboard, show a "Paste & Replace" button on the right with info bubble explaining the feature: "Replace this page's root component with the component copied from the clipboard."

### Tasks
- Allow root style copy/paste
- Copy/paste root component
- Feature to replace root component

abemscac:
Close #44

### DIFF
--- a/libs/frontend/feature-build/src/lib/use-copy-paste.ts
+++ b/libs/frontend/feature-build/src/lib/use-copy-paste.ts
@@ -23,12 +23,7 @@ export const useCopyPaste = (): IUseCopyPaste => {
   const pressCopy = (evt: KeyboardEvent) => {
     evt.stopImmediatePropagation()
     const component = editor.value?.selectedComponent
-    if (
-      site.value.editor &&
-      // Make sure selected component is not page root before copying
-      component?.parent &&
-      (evt.ctrlKey || evt.metaKey)
-    ) {
+    if (site.value.editor && component && (evt.ctrlKey || evt.metaKey)) {
       site.value.editor.copiedComponent = serializeComponent(component)
       // TODO -- find a good way to handle translations without requiring library consumers
       // to use `vue-i18n`
@@ -64,7 +59,12 @@ export const useCopyPaste = (): IUseCopyPaste => {
     const { selectedComponent } = editor.value ?? {}
     const { copiedComponent } = site.value.editor ?? {}
     if (selectedComponent && copiedComponent && (evt.ctrlKey || evt.metaKey)) {
-      if (evt.shiftKey) {
+      if (!copiedComponent.parentId) {
+        // Root component should not be pasted through paste hotkey.
+        addHUD({
+          text: 'Root component can only be used to replace another root component',
+        })
+      } else if (evt.shiftKey) {
         // Make sure we don't paste style to the same component
         if (selectedComponent.id !== copiedComponent.id) {
           pasteStyle(copiedComponent)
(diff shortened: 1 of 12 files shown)


libs/frontend/feature-command/src/lib/component/replace-page-root.ts
```diff
@@ -0,0 +1,39 @@
+import { setSelectedComponent } from '@pubstudio/frontend/feature-editor'
+import { resolveComponent } from '@pubstudio/frontend/util-builtin'
+import { IReplacePageRootData } from '@pubstudio/shared/type-command-data'
+import { ISite } from '@pubstudio/shared/type-site'
+import { addComponentHelper, undoAddComponent } from './add-component'
+import { applyRemoveComponent, undoRemoveComponentHelper } from './remove-component'
+
+export const applyReplacePageRoot = (site: ISite, data: IReplacePageRootData) => {
```
abemscac, line 8: Hi @sampullman 

In the current implementation, the new component (pasted component) will be added to the site with `addComponentHelper`, so the `nextId` in site context will be incremented during replace, and will be decremented (restored) when undo. Do you think this make sense?

libs/frontend/feature-command/src/lib/component/remove-component.ts
```diff
@@ -45,8 +45,17 @@ export const applyRemoveComponent = (site: ISite, data: IRemoveComponentData) =>
...
 export const undoRemoveComponent = (site: ISite, data: IRemoveComponentData) => {
+  undoRemoveComponentHelper(site, data, true)
+}
+
+// Declare a new function so that we don't have to add additional arguments to `undoRemoveComponent`.
+export const undoRemoveComponentHelper = (
+  site: ISite,
+  data: IRemoveComponentData,
+  selectAfterUndo: boolean,
+) => {
```
abemscac, line 58: I have to make a new function for undoing remove component because I need a new argument to determine if we should automatically select the restored component after undo.

The main reason is when undoing replace page root, old components (old root and its children) will be added back by using `undoRemoveComponent` command. The auto-select affected component feature in `undoRemoveComponent` will try to find which page does the restored component belong to, switch to that page & select that component. But at this moment, the root of the replaced page has not been restored yet (because the undo remove is executed bottom-up), so an error will be thrown. To avoid this problem I decided to make an almost identical function as the original `undoRemoveComponent`, but provide the third argument `selectAfterUndo` so that we don't have to add additional argument on `undoRemoveComponent`, which is used in `undoCommand` helper.

This sounds very complicated; I think I didn't explain it very well 😬. Feel free to let me know if you need further explanation!

libs/frontend/feature-build/src/lib/use-copy-paste.ts
```diff
@@ -70,7 +65,14 @@ export const useCopyPaste = (): IUseCopyPaste => {
           pasteStyle(copiedComponent)
         }
       } else {
-        pasteComponent(copiedComponent)
+        if (!copiedComponent.parentId) {
```
sampullman, line 68: This is probably good to do, but is there anything actually blocking us from pasting the root component into an arbitrary component? Maybe I'll test it out 🤔

libs/frontend/feature-command/src/lib/component/replace-page-root.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+        name: copiedComponent.name,
+        tag: copiedComponent.tag,
+        content: copiedComponent.content,
+        sourceId: copiedComponent.id,
+        parentId: '',
+      },
+    }
+  })
+
+  it('should replace page root with the specified component', () => {
```
sampullman, line 141: It could be useful to add a test where a new page's root is pasted to, but this sufficient for now.

libs/frontend/feature-command/src/lib/component/remove-component.ts
sampullman: It makes sense! I've wondered if selecting/unselecting components within commands is good, since it's the only "side effect" (non-`site.context` change), but I'm not sure how else to do it.

PR #138, opened 2023-12-09
fix: behavior modal javascript highlighting bug

### LINKED ISSUE #11: Behavior modal javascript highlight bug
The problems only occur when the text is being copied from a outer source (any place that's not the editor itself), and is pasted to the editor by using "ctrl/alt + a" -> "ctrl/alt + v" (select all then paste).

https://github.com/samatechtw/pubstudio-builder/assets/59676941/aa8d5df7-d2ec-4da7-b43d-e93b52ad215d

In the above video, new line characters are ignored after paste, and the edit history is cleared (undo/redo don't work anymore).

If the content is appended to the editor instead of replacing the whole content with "ctrl/alt + a" -> "ctrl/alt + v" , the editor will work fine.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/a1eb52ad-5915-40f8-a3aa-f9353e6c2da2

However, if the select all action is performed with mouse click or shift + arrow key, the problems will not occur.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/b7f35c85-0b2f-414f-99bd-b69763e96e3c

What's even more strange is, if the text is copied from a ProseMirror editor, pasting it with "ctrl/alt + a" -> "ctrl/alt + v" won't cause any problem. New line characters are preserved, and edit history still works.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/6f779838-a291-4dd9-84f3-0005280ad48c

From the log it seems that the selection produced by "ctrl/alt + a" and "mouse click/shift + arrow key" are programmatically different, even though they look the same on the UI. In short, the editor will broke when `newState = prosemirrorSetup(...)` is run. Maybe the only way to solve this problem is to modify the view state instead of creating a new one.

abemscac:
Close #11

### DIFF
--- a/libs/frontend/feature-build/src/lib/code-editor-plugins.ts
+++ b/libs/frontend/feature-build/src/lib/code-editor-plugins.ts
@@ -1,0 +1,56 @@
+import { schemaCode } from '@pubstudio/frontend/util-edit-text'
+import { Keys } from '@pubstudio/frontend/util-key-listener'
+import hljs from 'highlight.js/lib/core'
+import { keymap } from 'prosemirror-keymap'
+import { Fragment, Slice } from 'prosemirror-model'
+import { Plugin } from 'prosemirror-state'
+import { highlightPlugin } from './prosemirror-highlighthjs-plugin'
+
+// Manually create a new slice instead of using the third parameter `slice` from
+// `handlePaste` because new line characters will be ignored for unknown reason
+// when using it to do content replacement in the editor state.
+const transformPasteContentPlugin = new Plugin({
+  props: {
+    handlePaste: (view, e) => {
+      const pastedContent = e.clipboardData?.getData('text') ?? ''
+      if (pastedContent) {
+        const textNode = schemaCode.text(pastedContent)
+        const slice = new Slice(Fragment.from(textNode), 0, 0)
+        view.dispatch(view.state.tr.replaceSelection(slice))
+      }
+      return true
+    },
+  },
+})
+
+// Manually insert a new line instead of using the default behavior to avoid creating
+// a new block of <p> as a sibling of <pre>.
+const interceptNewLineEventPlugin = new Plugin({
+  props: {
+    handleDOMEvents: {
+      keydown: (view, e) => {
+        const { key, shiftKey, ctrlKey, metaKey } = e
+        if (key === Keys.Enter && (shiftKey || ctrlKey || metaKey)) {
+          // Use e.preventDefault() to prevent two new lines from being inserted when pressing
+          // Shift + Enter.
+          e.preventDefault()
+          view.dispatch(view.state.tr.insertText('\n'))
+          return true
+        }
+      },
+    },
+  },
+})
+
+export const codeEditorPlugins: Plugin[] = [
+  keymap({
+    // Insert 2 spaces in the selected position when tab is pressed.
+    Tab: (state, dispatch) => {
+      dispatch?.(state.tr.insertText('  ').scrollIntoView())
+      return true
+    },
+  }),
+  highlightPlugin(hljs),
+  transformPasteContentPlugin,
+  interceptNewLineEventPlugin,
+]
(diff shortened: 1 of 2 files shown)


libs/frontend/feature-build/src/lib/code-editor-plugins.ts
```diff
@@ -0,0 +1,56 @@
...
+import hljs from 'highlight.js/lib/core'
+import { keymap } from 'prosemirror-keymap'
+import { Fragment, Slice } from 'prosemirror-model'
+import { Plugin } from 'prosemirror-state'
+import { highlightPlugin } from './prosemirror-highlighthjs-plugin'
+
+// Manually create a new slice instead of using the third parameter `slice` from
+// `handlePaste` because new line characters will be ignored for unknown reason
+// when using it to do content replacement in the editor state.
+const transformPasteContentPlugin = new Plugin({
```
abemscac, line 12: Hi @sampullman 

I think it's finally solved! For some reason, new line characters from an outer source (not copied within a prosemirror editor) will be ignored when pasted to to editor. I'm not sure if it's something related to how we use the prosemirror editor or how we define the schema 🤔?

Manually create a new slice and inserting it back to the editor seems to work here.

libs/frontend/feature-build/src/lib/code-editor-plugins.ts
```diff
@@ -0,0 +1,58 @@
...
+      return true
+    },
+  },
+})
+
+// Manually insert a new line instead of using the default behavior to avoid creating
+// a new block of <p> as a sibling of <pre>.
+// Use handleDOMEvents insead of keymap because combinations like 'Shift-Enter' didn't
+// work for some reason; our callback function never gets executed.
+const interceptNewLineEventPlugin = new Plugin({
```
abemscac, line 30: I noticed that pressing Cmd + Enter, Ctrl + Enter, or Shift + Enter will create an "incorrect" new line in the editor. I have to add this event handler plugin so that those key combinations will lead to a normal new line just like what Enter does.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/692b79ea-50f8-454d-ae37-670c6ee5cc90

I use `handleDOMEvents` here insead of keymap because for some reason, combinations like 'Shift-Enter' didn't
work in keymap 😕. The callback function gets called.

libs/frontend/feature-build/src/lib/create-code-editor-view.ts
```diff
@@ -42,21 +30,20 @@ export function createCodeEditorView(
...
-        //    Find a way to replace the content in the editor state while keeping everything.
-        // 2. New line characters are ignored if the text is copied from a source other than ProseMirror
-        //    editor. For example, VSCode.
-        newState = prosemirrorSetup({
-          schema: schemaCode,
-          content: `<pre><code>${textContent}</code></pre>`,
-          plugins: [...codeEditorPlugins],
-        })
+        // feature to stop working. So we need to manually add the `code_block` node back in such cases.
+        const codeBlockNode = schemaCode.node(
```
abemscac, line 38: This strange workaround is also fixed! I found a way to create a new `code_block` node & insert it back to the editor content so that we don't have to initialize a brand new editor state like we did before. I think edit history (redo/undo) should be working fine after" ctrl + a -> ctrl + v" now!

libs/frontend/feature-build/src/lib/code-editor-plugins.ts
sampullman: Interesting! I'm not sure why the keymap doesn't work.

libs/frontend/feature-build/src/lib/create-code-editor-view.ts
```diff
@@ -42,21 +30,20 @@ export function createCodeEditorView(
...
+          'code_block',
+          undefined,
+          textContent ? schemaCode.text(textContent) : undefined,
```
sampullman, line 41: Looks good overall! I made this one change, because apparently ProseMirror doesn't allow empty text nodes. Previously, selecting everything and pressing delete would cause a crash.

PR #140, opened 2023-12-11
fix: component tree child padding bug

### LINKED ISSUE #132: Component tree child padding bug
When a component tree item's children are all collapsed, their padding disappears.

I updated how the component tree styles work a couple weeks ago, this bug is likely due to how the tree depth is calculated.

https://github.com/samatechtw/pubstudio-builder/assets/550982/1c3e6524-319c-40ea-ad5d-17075feebb06

abemscac:
Close #132

### DIFF
--- a/libs/frontend/ui-build/src/components/ComponentTree.vue
+++ b/libs/frontend/ui-build/src/components/ComponentTree.vue
@@ -21,18 +21,11 @@ const root = computed(() => activePage.value?.root)
 
 const show = computed(() => editor.value?.showComponentTree)
 
-interface IterNode {
-  cmp: IComponent
-  depth: number
-}
-const makeNode = (component: IComponent, depth: number): IterNode => ({
-  cmp: component,
-  depth,
-})
-
 const levelStyles = computed(() => {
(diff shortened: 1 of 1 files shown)


libs/frontend/ui-build/src/components/ComponentTree.vue
```diff
@@ -44,25 +37,27 @@ const treeDepth = computed(() => {
...
+    for (let i = 0; i < length; i++) {
+      const cmp = queue.shift() as IComponent
+      if (expandedCmp[cmp.id]) {
+        cmp.children?.forEach((child) => queue.push(child))
       }
     }
+    depth++
   }
+
   return depth
```
abemscac, line 56: Hi @sampullman 

I change the way to calculate tree depth from using a stack to using a queue. There's not much of a difference, but I think it makes it easier to fix the padding issue. I hope this is okay 😬.

sampullman: No problem!

sampullman:
Looks good to me!

PR #142, opened 2023-12-12
feat: link tooltip improvements

### LINKED ISSUE #134: Link tooltip improvements
Implement the following improvements/fixes for the link tooltip, which appears in the builder when a Link component is selected.

1. Show the tooltip to the left and/or bottom of the component instead of overflowing, when the component is at the top or right of the builder. This could be done with `floating-ui`, or manually by checking positions when the component is selected.
2. Reuse the datalist shown in the inputs/href edit, so the tooltip can easily select pages/locations as targets.
3. The text color of the link/input is influenced by the component styles, it should be the default site text color when not editing, and black when editing.

abemscac:
Close #134


libs/frontend/feature-render-builder/src/components/ProseMirrorEditor.ts
```diff
@@ -203,7 +203,7 @@ export const ProseMirrorEditor = defineComponent({
 
     return () => {
       return h('div', {
-        id: containerId,
+        id: containerId.value,
```
abemscac, line 206: I accidentally saw an element with id `[object Object]` in the inspector, and this seems to fix it.

libs/frontend/util-tooltip/src/lib/use-tooltip.ts
```diff
@@ -50,6 +52,9 @@ export const useTooltip = (options?: IUseTooltipOptions): IUseTooltip => {
   if (options?.shift) {
     middleware.push(shift())
   }
+  if (options?.flip) {
+    middleware.push(flip())
```
abemscac, line 56: I added a `flip` option that uses the [flip](https://floating-ui.com/docs/flip) middleware so that tooltip would flip up/down when needed.
However, the tooltips (edit/copy) inside the LinkTooltip will cause a "temporary" overflow for some reason.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/cd75153f-aede-4fc3-860b-f9ab8d0f146c

It may not be obvious, but at the beginning there's a small horizontal overflow happening in the builder. After swiping left and right with touchpad for a couple times, the overflow just vanished all of a sudden! Vertical swipes (that is, scroll up and down) didn't have the same effect.

I'm not 100% sure why this is happening; maybe those horizontal swipes triggered the auto-update behavior we wrote in `use-tooltip` composable.

The [autoPlacement](https://floating-ui.com/docs/autoplacement) middleware did fix this, but it also led to another problem - the edit&copy tooltip is now always positioned to the left/right side of the item. We also have to manually deal with the position&rotation of the arrow.

Screenshot 2023-12-12 at 5 39 00 PM

I think this happens because the component ("Pricing" link) is the reference element of these tooltips, and there was't enough vertical space for the tooltips, so it just went horizontal.

It would be great if you could take a look at this, sorry about that! 😵

libs/frontend/feature-render-builder/src/components/ProseMirrorEditor.ts
sampullman: Good catch 😬

libs/frontend/util-tooltip/src/lib/use-tooltip.ts
sampullman: I see what you mean! But I'm not able to reliably reproduce it 🤔 

I think the behavior is good enough for now, I don't see it causing a big problem.

libs/frontend/feature-render-builder/src/components/LinkTooltip.vue
```diff
@@ -1,18 +1,23 @@
...
       :draggable="true"
       @dragstart.stop.prevent
       @keyup.enter="updateLink"
+      @keyup.esc="$event.srcElement.blur()"
     />
     <div v-else-if="internalLink" class="link-view" @click="gotoPage">
       {{ link }}
     </div>
-    <a v-else-if="link" :href="link" target="_blank" class="link-view"> {{ link }}</a>
+    <a v-else-if="link" :href="link" target="_blank" class="link-view" @click.stop>
```
sampullman, line 18: @abemscac I was able to get rid of the workaround by adding `@click.stop` here. I think there's something overriding the click behavior, so this makes sure the link works.

PR #145, opened 2023-12-14
135 UI consistency improvements

### LINKED ISSUE #53: Preview CSS changes
Allow users to preview CSS changes when editing. When a component CSS value is edited in the right menu, or modified in the text/bg color menu

If the user exits the CSS editing mode, revert the change. The result should be that a command is not committed for the change.

This behavior may be confusing if a user sets a color and exits the color picker, so we may want to commit the change when clicking outside, and only revert the change if "Cancel" is clicked in the picker.

### LINKED ISSUE #135: UI consistency and improvements
There are a few places where the UI is inconsistent or clunky, and can be improved.

- [x] Left menus that don't cover the component tree should have a shadow
- [ ] Custom style editing is awkward because the inputs are often too small for the values. Consider switching to 2 rows while editing, or other ways to improve.
- [x] Remove the debug button or limit to dev
- [x] Consider custom scrollbar for component menu/editor, to reduce space
- [x] Improve inputs UI
- [x] Move Delete/Clear site from toolbar to Site Settings
- [x] Move breakpoints to main toolbar instead of dropdown
- [ ] Experiment with color scheme, consider light/dark

sampullman:
Close #135 

Close #53

WIP - component style editing is functional, but refactoring is needed to support live editing for mixins and child style overrides. In theory, `useEditComponentStyles` should be mostly reusable, but some effort is needed to make it work with the mixin/override commands. It might be simpler to copy/paste `useEditComponentStyles` and adjust, instead of reusing.


sampullman:
@abemscac I had to do a lot more refactoring than I hoped to even get the feature working for component custom styles, but I feel a little more confident about it now. I'll leave a couple comments around the most significant changes.

I originally implemented the CSS preview feature for child styles/overrides and mixins as well, but my approach was messy and there were too many bugs. I rolled it back, so this branch only includes component custom styles - I'll create a new issue for including overrides/mixins.

I'm sure there are still bugs here and there, but I think it's better to merge now and find them through testing rather than spend even more time trying to make it perfect.

libs/frontend/feature-build/src/lib/use-edit-component-styles.ts
```diff
@@ -0,0 +1,224 @@
...
+        | ICommand<ISetComponentCustomStyleData>
+        | undefined
+    }
+    const originalStyle = removeCmd?.data.oldStyle
+      ? { ...removeCmd.data.oldStyle }
+      : undefined
+    return { removeCmd, originalStyle }
+  }
+
+  const setStyle = (
```
sampullman, line 124: This is pretty much the whole "preview" feature. It can probably be cleaned up a little more, but the idea is:
- When a style is updated, check for a command that affects the target property in the `editCommands` (used to track active style editing commands)
- If found, remove the command and replace it with the new one
- Replace the command in history, or add a new command on the first edit

libs/shared/type-site/src/lib/i-editor-context.ts
```diff
@@ -90,6 +91,8 @@ export interface IResizeData {
 }
 
 export interface IEditorContext {
+  // Reference to store used to save/restore the site, populated on initialization
+  store?: ISiteStore
```
sampullman, line 95: I spent a while trying to make all the dependencies work, but in the end I think it's worth including a reference to the store in the editor context. It's really easy to set up, and it means we can avoid importing the source/store in many places. I hope it makes sense.

abemscac:
Got it! I'll take a look at and try to understand everything.
Thank you!

Issue #149, opened 2023-12-16
Color picker for text not working


sampullman:
Recently, setting the text or background color of a text component doesn't work when the ProseMirror editor is not active.

**Current behavior:** Color not updating and ProseMirror editor is activated

**Expected behavior:** Color is updated and editor is not activated

PR #152, opened 2023-12-17
fix: further bug fixes for toolbar text styling #149

sampullman:
Related #149


libs/frontend/feature-build/src/lib/use-toolbar-resume-text-focus.ts
```diff
@@ -0,0 +1,37 @@
+import { IEditorContext } from '@pubstudio/shared/type-site'
+import { Ref, ref } from 'vue'
```
sampullman, line 2: @abemscac Here's another update worth noting! There were some problems with detecting whether to resume focus/apply toolbar styles to a ProseMirror text selection, so I added this hack/workaround after struggling for a while. The idea is to detect whether the ProseMirror editor was focused on `mousedown` of the toolbar button, then use that information once a new style is applied, to re-focus the editor.

It seems to work well for the color picker and font size so far. I thought it would be necessary for bold/italic/font family as well, but it seems like the previous way of re-focusing the editor is good enough. I should probably look into it better and document somewhere, because I'm sure I'll be confused about it again in the future.

abemscac: Got it, thanks for letting me know!

PR #154, opened 2023-12-18
feat: ui for setting link target _blank

### LINKED ISSUE #124: UI for setting link target __blank
Add a checkbox for toggle the link's target to/from `__blank`

The checkbox label should be "Open in new tab".

If the target is set to something other than `__blank`, it should appear unchecked, and setting/pressing ctrl z will result in the old `target`.

abemscac:
Close #124

### DIFF
--- a/libs/frontend/feature-render-builder/src/components/LinkTooltip.vue
+++ b/libs/frontend/feature-render-builder/src/components/LinkTooltip.vue
@@ -21,7 +21,17 @@
     <div v-else class="empty">
       {{ t('build.no_link') }}
     </div>
-    <Check v-if="editing" class="link-save" color="#009879" @click="updateLink" />
+    <div v-if="editing" class="editing-wrap">
+      <Checkbox
+        class="open-in-new-tab-checkbox"
+        :item="{
+          label: t('build.open_in_new_tab'),
+          checked: openInNewTab,
+        }"
+        @checked="openInNewTab = $event"
+      />
+      <Check class="link-save" color="#009879" @click="updateLink" />
+    </div>
     <div v-else class="icon-wrap">
       <IconTooltip :tip="t('edit')">
         <Edit class="edit-link" @click="edit" />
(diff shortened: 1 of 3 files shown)


libs/frontend/feature-render-builder/src/components/LinkTooltip.vue
```diff
@@ -56,36 +75,64 @@ const { link, componentId } = toRefs(props)
...
 const gotoPage = () => {
   changePage(link.value)
 }
 
 const edit = async () => {
   editing.value = true
   editedLink.value = link.value
+  openInNewTab.value = componentTargetInput.value === '_blank'
   await nextTick()
   linkInput.value?.inputRef?.focus()
```
abemscac, line 101: Hi @sampullman 

The URL input in LinkTooltip gets auto-focused when edit button is clicked. This worked well before, but after anchors/pages datalist has been added to the input, the datalist is now covering the other UI (open in new tab checkbox and save button) when the list is too long. Even if we disable the auto-focus feature, the list will still cover the UI when user manually focus the input as long as there're enough items in the list.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/6deae992-04db-4760-8ad8-8c3c02b24202

Do you have an idea how can we improve this 🤔?

sampullman: I experimented a bit and couldn't get `datalist` styling to work as expected. I think we'll either need to use/adapt `PSMultiselect`, or maybe move the checkbox to the left side as a temporary improvement.

I'll merge this PR now though, because I don't think it's a critical problem. Selecting an item or pressing enter closes the datalist, which is ok.

abemscac: I tried to set the style of datalist with CSS yesterday, but it seems that the style of datalist is not customizable. I wasn't able to replace datalist with PSMultiselect in time, sorry about that! I'll include the change in another PR instead, thanks!

sampullman: Got it, no problem!

PR #169, opened 2023-12-22
feat: implement links in prosemirror editor

### LINKED ISSUE #126: Implement links in ProseMirror editor
Implement inline links for the component/translations ProseMirror editor.

If it's possible to reuse the `LinkTooltip` component we should do so, and if possible include a checkbox for opening in a new tab (set `target="__blank"`).

abemscac:
Close #126


libs/frontend/feature-render-builder/src/components/LinkTooltip.vue
```diff
@@ -117,8 +145,30 @@ const updateLink = async () => {
...
+        // `nodesBetween` will gives us both root node and link node.
+        // Check `childCount` to make sure it is the link node that gets replaced.
+        if (node.childCount === 0) {
+          const endPos = startPos + node.nodeSize
+          const target = openInNewTab.value ? '_blank' : undefined
+          // TODO: is it necessary & possible to keep other marks? (color, font-weight, font-family, etc)
+          const newLinkNode = createLinkNode(schemaText, editedLink.value, target)
+          editView.value?.dispatch(state.tr.replaceWith(startPos, endPos, newLinkNode))
+        }
+      })
```
abemscac, line 169: Hi @sampullman 

The use of `state.doc.nodesBetween()` here is to find & replace the node at the current cursor. It worked, but it seems awkward 😬. I think there should be a better way!

libs/frontend/feature-render-builder/src/components/ProseMirrorEditor.ts
```diff
@@ -201,12 +205,44 @@ export const ProseMirrorEditor = defineComponent({
...
+        {
+          id: containerId.value,
+          ref: container,
+          class: 'prose-mirror-editor-container',
+        },
+        cursorLinkAttrs.value
+          ? h(
+              // Use teleport to prevent LinkTooltip from constantly changing position
+              // due to the relative/absolute layout in the builder.
+              Teleport,
```
abemscac, line 230: Use teleport here to prevent `LinkTooltip` from flying around due to the absolute/relative layout in the builder (i.e. `force-relative` class).

https://github.com/samatechtw/pubstudio-builder/assets/59676941/16703b58-3824-43e2-bd09-680c77550149

libs/frontend/util-edit-text/src/lib/keymap.ts
```diff
@@ -109,5 +110,10 @@ export function buildKeymap(schema: Schema, overrideKeys?: { [key: string]: Comm
     })
   }
 
+  if ((type = schema.nodes.text)) {
+    bind('Space', textToLinkOnSpace(schema))
+    bind('Enter', textToLinkOnEnter(schema))
+  }
```
abemscac, line 116: Hi @sampullman 

I think the reason why Enter is not working in our custom keymap is because [Enter](https://github.com/ProseMirror/prosemirror-commands/blob/2da5f6621ab684b5b3b2a2982b8f91d293d4a582/src/commands.ts#L711C12-L711C25) is already handled in the baseKeymap from `prosemirror-commands`. It returns true in most cases, so any keymap added after that will not get executed.

Here's a list of all keys handled in the base keymap:

Screenshot 2023-12-22 at 5 26 16 PM

libs/frontend/util-edit-text/src/lib/prosemirror-utils.ts
```diff
@@ -48,7 +49,19 @@ export const firstMarkInSelection = (
...
   const { $cursor, ranges } = selection
   if ($cursor) {
-    return mark.isInSet(state.storedMarks || $cursor.marks())
+    const marks = $cursor.marks()
+    const markInSet = mark.isInSet(state.storedMarks || marks)
+    if (markInSet) {
+      return markInSet
+    } else {
+      // FIXME: this is a workaround for link mark not being found in `mark.isInSet`.
+      const firstMark = marks[0]
```
abemscac, line 58: I'm not sure why `firstMarkInSelection` always returns false for link mark 🤔.
So I made a temporary workaround that check `mark.type.name`.

abemscac:
Hi @sampullman 

I think the inline link should work as expected in most cases, but I noticed if the URL does not starts with http(s), the link would be considered as relative in preview & live mode. For example. `www.google.com` would become `localhost:8080/preview/www.google.com` in development mode.

Screenshot 2023-12-22 at 5 29 48 PM

Does this mean we should force users to make all inline links start with either http or https 🤔?

sampullman:
I'll think about it! For now, I think we can accept the current behavior. Some options for improving could be detecting if a link is valid for the current sight, and either warning the user about invalid links or automatically adding `https`

libs/frontend/util-edit-text/src/lib/keymap.ts
sampullman: Yeah, that makes sense. So to have our own custom handler, it's probably easiest to copy/paste from `[prosemirror-keymap](https://www.npmjs.com/package/prosemirror-keymap)` (or wherever it's defined), and modify it to include our own code.

libs/frontend/util-edit-text/src/lib/prosemirror-setup.ts
```diff
@@ -12,14 +12,27 @@ export interface IProsemirrorSetupOptions {
...
+  const baseKeyBinding = { ...baseKeymap }
+  overwriteBaseKeys?.forEach((key) => {
+    delete baseKeyBinding[key]
+  })
+
   const plugins = [
     buildInputRules(schema),
     keymap(buildKeymap(schema, mapKeys)),
-    keymap(baseKeymap),
+    keymap(baseKeyBinding),
```
abemscac, line 35: I just realized the keymap for Space and Enter would influence code editor as well if they're placed inside `buildKeymap`! So I remove them from `buildKeymap`, make them a new plugin and only use it in `createComponentEditorView`. But I have to add a new option `overwriteBaseKeys` here to prevent several keys from being handled by `baseKeymap` (i.e. Enter).

sampullman:
@abemscac Sorry for the delay! I reviewed the functionality over the weekend, I meant to take a closer look at the code before commenting but have been stuck on another issue.

I have two general comments:

1. Clicking a link created in the ProseMirror editor when the component isn't active navigates to the link destination. I'm not sure what the best/easiest solution is - ideally it would activate prosemirror and select the link, but if that's difficult it's ok to just activate the editor.
2. The link tooltip shows above or below the editor, instead of next to the link itself. This is ok for single line text components, but doesn't work well for multiple lines.

abemscac:
Got it! I'll change it accordingly.
Thanks!

abemscac:
Hi @sampullman 

I think the tooltip position is solved; it'll always show up around the target link instead of on the top/bottom of the selected component. Unfortunately, I didn't manage to set the cursor of a ProseMirror editor to the clicked link when a component is not selected, but here are some findings that might be useful in the future:

- The `<a>`s rendered when a component is/isn't selected are different (`===` operator would return `false`). If we were to find the clicked link (element) after the ProseMirror editor is mounted, we could use [`Node.isEqualNode`](https://developer.mozilla.org/en-US/docs/Web/API/Node/isEqualNode) to check if two nodes/elements are "the same".
- [posAtDOM](https://prosemirror.net/docs/ref/#view.EditorView.posAtDOM) and [posAtCoords](https://prosemirror.net/docs/ref/#view.EditorView.posAtCoords) might be the method we need to get ProseMirror node position by DOM element.
  - `posAtDOM` didn't work because it always return `-1`. Maybe I got something wrong.
  - `posAtCoords` didn't work because it always return `{ pos: -1, inside: -1 .}`. Maybe I got something wrong.
- I also found something called [Node.compareDocumentPosition](https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition), which is meant to be used together with bitmask like [Node.DOCUMENT_POSITION_CONTAINED_BY](https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition#node.document_position_contained_by), to check if a node/element is a descendant of another node/element. That being said, the result is always `false` when used on ProseMirror editor element & the clicked link, and on the corresponding paragraph (`<div class="pm-p">`) & the clicked link as well.
  - I'm not sure if it's the same thing as [Node.contains](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains); they seemed to do the same thing to me.

libs/frontend/util-edit-text/src/lib/prosemirror-setup.ts
sampullman: Makes sense, although it feels like there should be an easier way! Maybe in the future we can copy in the `baseKeymap` definition from prosemirror and use it in the `buildKeymap`. That way overriding would work as expected.

libs/frontend/util-edit-text/src/lib/prosemirror-utils.ts
sampullman: I see! I'll try to remember to investigate this, but probably won't have time for a while. It takes me a while to solve anything ProseMirror related, I feel like I have to relearn things every time I look at it 😅

libs/frontend/util-edit-text/src/lib/prosemirror-utils.ts
```diff
@@ -142,3 +171,12 @@ export const replaceMark = (markType: MarkType, attrs: Attrs | null = null): Com
...
   }
 }
+
+export const createLinkNode = (schema: Schema, href: string, target?: string) => {
+  const attrs = {
+    title: href,
+    href: href,
+    target: target ?? null,
+  }
+  return schema.text(href, [schema.mark('link', attrs), schema.mark('u')])
```
sampullman, line 181: Is there a difference between `schema.mark('link')` and `schema.marks.link.create()`?

libs/frontend/util-edit-text/src/lib/prosemirror-setup.ts
abemscac: Yeah, I actually make a copy of `baseKeymap` for #173.
Is it okay if I include the fix for #173 in this PR?

libs/frontend/util-edit-text/src/lib/prosemirror-utils.ts
abemscac: Yes, I think so! It's just that `schema.mark('link')` would throw an error immediately if "link" is not defined in the schema because it automatically calls `create()`, while `schema.marks.link` will be `undefined`, so no error will be thrown as long as we don't call `.create()`.

libs/frontend/util-edit-text/src/lib/prosemirror-setup.ts
sampullman: Sure! I have changes here but I think it will be easy to rebase.

abemscac: Got it! Do you wish to push your changes first? Or is it going to be easier for you if I push mine first 🤔?

sampullman:
I think it's better for you to push first, mine are easy to recreate if merging is too difficult.

libs/frontend/util-edit-text/src/lib/base-keymap.ts
```diff
@@ -0,0 +1,995 @@
...
+          if (
+            deflt &&
+            $from.node(-1).canReplaceWith($first.index(), $first.index() + 1, deflt)
+          )
+            tr.setNodeMarkup(tr.mapping.map($from.before()), deflt)
+        }
+      }
+      // Originally this was `dispatch(tr.scrollIntoView())`. `scrollIntoView()` was removed
+      // owing to https://github.com/samatechtw/pubstudio-builder/issues/173.
+      dispatch(tr)
```
abemscac, line 529: Hi @sampullman 

This file is a copied of the `baseKeymap` from `prosemirror-commands`. But I have to make an adjustment here to solve the bug found in #173.

libs/frontend/util-edit-text/src/lib/text-to-link-keymap.ts
```diff
@@ -0,0 +1,103 @@
...
+          // Pass the updated state to base command so that transaction could work properly.
+          baseKeymap.Enter?.(newState, dispatch, view)
+        }
+        return true
+      } else {
+        return chainCommands(createParagraphNear, liftEmptyBlock, splitBlock)(
+          state,
+          dispatch,
+          view,
+        )
```
abemscac, line 61: The base `Enter` keymap is `chainCommands(newlineInCode, createParagraphNear, liftEmptyBlock, splitBlock)`, but since this plugin is only used in ProseMirrorEditor for component, I removed `newlineInCode` here and use the commands from the copied version of `baseKeymap` for #175 .

libs/frontend/util-edit-text/src/lib/prosemirror-setup.ts
abemscac: Hi @sampullman 

I've pushed the fix for #173 a couple minutes ago. There's a lint error related to the use of `os` in `base-keymap.ts`. I think `os` is only available in NodeJS, so I wonder if we could just comment out that part for now🤔?

sampullman: No problem, I saw that as well. I'll merge my changes and double check everything, then remove the `os` usage!

libs/frontend/feature-render-builder/src/components/LinkTooltip.vue
sampullman: I think `nodesBetween` works, but I'm not sure `node.childCount === 0` always works if there are overlapping nodes.  It may better to check which node has a `link` mark present. Another thing you can do to get the node is `$cursor.parent.child($cursor.index())`.

While thinking about this I noticed if there is another mark within the link text area (like text color), only that inner node text will be replaced with the above approach. So changing `https://google.com` to `https://test.com` might result in `https://https://test.com.com`, if the `google` part is styled. My hacky solution is to check nodes before/after the current one to see if they have a link mark, and adjust the start/end replacement position.

For retaining existing marks, I would need to test a few things, but I think it's possible in a limited way. We can add some code to check each node we're replacing, and if the new link has the same substring (at the same start position), skip it instead of replacing. However, I think this is a lot more effort than it's worth right now.

Finally, I've been considering how to do things like "style all links the same way automatically". I think one option is to have some special theme variables that we resolve when links are created, and apply them to the mark. For example `${link-color}`, `${link-visited-color}`, etc. It's a low priority feature, though!

libs/frontend/feature-render-builder/src/components/ProseMirrorEditor.ts
```diff
@@ -201,12 +205,51 @@ export const ProseMirrorEditor = defineComponent({
       },
     )
 
+    const cursorLinkAttrs = ref()
+    let linkElement: Element | undefined = undefined
+
+    watch(editViewTxCount, () => {
+      const { editView } = editor.value
```
sampullman, line 212: This might end up being inefficient, I wonder if we can specifically observe whether the selection changed, by setting a different variable in `dispatchTransaction`. I'll add a note to check it eventually!

libs/frontend/feature-render-builder/src/components/ProseMirrorEditor.ts
sampullman: Sounds good! I moved the Teleport to `LinkComponent`, because I think it's fine to include in both situations (prosemirror and builder links).

PR #180, opened 2023-12-28
feat: open reusable style page after undo

### LINKED ISSUE #159: Open reusable style page after undo
When undoing a command that edits a reusable style, the editor right menu should show the style editing menu.

When undoing a command while the reusable style editing menu is open, and the previous command was not related to reusable styles, the menu should be hidden. Basically, the right menu should return to the correct context during undo/redo.

abemscac:
Close #159 

Hi @sampullman 

Sorry, I kept mixing up undo/redo 😵.
It would be great if you could double check whether the added code make sense. My brain is malfunctioning so hard at this issue. I apologize for any recent inconvenience caused!


sampullman:
No problem, I'll take a look today!

sampullman:
Due to how I modified the edit mixin commands, this is even more complicated that I expected!

I've made some modifications to this branch, but I'm still not convinced about the right solution. I mentioned that I want to avoid adding commands that only affect the editor, but now I'm wondering if it's necessary here.

I'm going to think about it a little more while continuing to review templates, but I'm going to try the following:
- Merge the edit-mixin name command with the rest of the command group at the end of a mixin edit "session"
- Add a special command `set-editor-mixin-menu` to remember the mixin edit state
- Add `set-editor-mixin-menu` to the command stack when the edit session ends. On undo/redo it opens/closes the menu.

I think this will make undoing a mixin command feel more natural, and should clean up the command code a bit, but it doesn't cover the problem of undoing several commands, then redoing the mixin edit command 🤔

sampullman:
@abemscac I followed through with my comment and I think it works, although it took some refactoring!

I'll try to test it more extensively while working on templates.

PR #186, opened 2023-12-30
fix: live mode mailto and other link bugs #184

### LINKED ISSUE #184: mailto links don't work in preview/live
Links with `mailto` are treated as relative URLs in preview and live render modes.

sampullman:
Close #184


libs/frontend/feature-render/src/lib/live-component.ts
```diff
@@ -65,17 +65,15 @@ export const LiveComponent = () => {
...
           const path = props.href ?? ''
-          return h(
-            RouterLink,
-            {
-              ...renderProps,
-              to: { path },
-            },
-            () => children,
-          )
+          const linkProps = {
```
sampullman, line 70: @abemscac This was harder than I expected, there's some interaction between Vue props and the root child props in a component that I don't quite understand. I either had to set the correct `href` here, or delete it to let the component handle it, otherwise it would be overridden.

For example, I added code in `RouterLink` to transform `www.site.com` to `https://www.site.com`, but since the `href` was present in the props here, it took priority and the relative url was used. I'm not sure if this is the case for all props, or only ones that appear in the HTML.

PR #188, opened 2023-12-31
187 template site

### LINKED ISSUE #168: Site name with special characters results in invalid Site ID
Sites names with spaces, pluses, cause problems since the resulting site ID is used in CSS.

Consider ways to solve this problem, such as:
- Convert or remove special characters when generating the Site ID. We would also want to check the API to make sure the ID is still unique.
- Don't allow invalid characters in Site name
- Allow the user to specify an ID, with only  `[a-z_-]` (still need to check for uniqueness

A combination of the above approaches could work, for example only allowing `([a-zA-z]|\s)` in site names, and converting spaces to underscores or dashes. This should solve the uniqueness problem as well.

### LINKED ISSUE #187: Set Site ID when template is loaded
When a template is loaded, all IDs should be replaced with the current site's ID.

Add a unit-tested helper that takes a site and new ID, and replaces all the IDs in the site with the new one.

- Set namespace
- Component IDs
- Mixin IDs
- Behavior IDs
- All ID references (inputs, component mixins, component ID, componentTreeExpandedItems

sampullman:
Close #187 

Close #168

### DIFF
--- a/libs/frontend/util-render/src/lib/util-render.ts
+++ b/libs/frontend/util-render/src/lib/util-render.ts
@@ -8,6 +8,7 @@ import {
   ISiteContext,
   IThemeFont,
 } from '@pubstudio/shared/type-site'
+import { breakpointSortFn } from './sort-breakpoint-fn'
 import { IQueryStyle, IRawStyleRecord } from './util-render-types'
 
 export const rawStyleRecordToString = (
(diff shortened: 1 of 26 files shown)


libs/frontend/util-render/src/lib/util-render.ts
```diff
@@ -52,15 +53,25 @@ export const queryStyleToString = (
...
-  Object.entries(context.breakpoints).forEach(([breakpointId, breakpoint]) => {
-    const rawStyle = rawStyleRecordToString(queryStyle[breakpointId], context)
-    if (breakpointId === DEFAULT_BREAKPOINT_ID) {
-      styleContent += rawStyle
-    } else {
-      const mediaQuery = getMediaQuery(breakpoint)
-      styleContent += `@media ${mediaQuery} {${rawStyle}}`
-    }
-  })
+  Object.entries(context.breakpoints)
```
sampullman, line 56: @abemscac I had to make this change for the styles to be correctly ordered in preview/live mode. Otherwise, some of my sites would show correctly in the editor, but the smaller breakpoint styles would be overridden by the default/larger styles.

This can be optimized a lot, for example the `reverse` isn't necessary, but I was busy with other things. I also wonder if it's just hiding another bug, since I thought this worked before. It might have only happened after editing breakpoints, I'm not sure.

sampullman:
@abemscac @ClaireMao1996 After this PR, you may need to run `rustup update` to get the latest version of Rust locally! Although it will probably run fine on the old version still.

PR #192, opened 2024-01-01
fix: prosemirror link creation removes text bug #185

### LINKED ISSUE #185: ProseMirror link creation removes text
When there's a link in a text component that hasn't been transformed yet, and there is more text after the link, pressing space/enter at the end of the text causes a bug.

Pressing enter creates the link, but deletes all the text between the link and the end of the component.

Pressing space causes an error to be thrown:

> Uncaught RangeError: Position 228 out of range

https://github.com/samatechtw/pubstudio-builder/assets/550982/782887e3-0278-4867-ae0d-b05700877c73

sampullman:
Close #185

### DIFF
--- a/libs/frontend/util-edit-text/src/lib/text-to-link-keymap.ts
+++ b/libs/frontend/util-edit-text/src/lib/text-to-link-keymap.ts
@@ -71,14 +71,14 @@ const convertTextToLinkWhenNecessary = (
   $cursor: ResolvedPos,
 ): Transaction | undefined => {
   const nodeBefore = $cursor.nodeBefore
-  const nodeBeforeIsLink = nodeBefore?.marks.some(
-    (mark) => mark.type.name === schema.mark('link')?.type.name,
-  )
+  const linkMark = schema.mark('link')?.type.name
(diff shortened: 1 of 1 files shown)


libs/frontend/util-edit-text/src/lib/text-to-link-keymap.ts
```diff
@@ -71,14 +71,14 @@ const convertTextToLinkWhenNecessary = (
...
-  )
+  const linkMark = schema.mark('link')?.type.name
+  const nodeBeforeIsLink = nodeBefore?.marks.some((mark) => mark.type.name === linkMark)
 
   if (!nodeBeforeIsLink && nodeBefore?.isText && nodeBefore.text) {
     // Convert text to link
     const match = urlRegex().exec(nodeBefore.text)
-    if (match) {
+    // Only convert when the link is the last text entered
+    if (match && nodeBefore.text.endsWith(match[0])) {
```
sampullman, line 81: @abemscac I didn't want to duplicate the regex to make it only match links at the end of the text, so I think this makes sense, even if it feels a little hacky. At least, it seems to fix the problem.

PR #195, opened 2024-01-02
fix: mixin breakpoint styles not showing in editor

### LINKED ISSUE #161: Scroll to selected component on editor load
When the editor page loads, scroll to the currently selected component.

### LINKED ISSUE #182: Mixin breakpoint styles not showing in editor
Editing reusable styles in non-default breakpoints appears to work correctly, but the resulting styles aren't displayed in the editor.

In preview and live mode, the styles are shown correctly.

https://github.com/samatechtw/pubstudio-builder/assets/550982/33d7d0bd-cfdd-44da-acb2-6c97e1e1b98e

abemscac:
Close #182 
Close #161


abemscac:
Hi @sampullman 

#161 seems like an easy one, so I included it here as well. I hope that's okay!

libs/frontend/feature-render-builder/src/lib/use-render-builder.ts
```diff
@@ -1,16 +1,54 @@
...
+  activeBreakpoint,
+  descSortedBreakpoints,
+} from '@pubstudio/frontend/feature-site-source'
+import { Css } from '@pubstudio/shared/type-site'
 
 export const useRenderBuilder = (options: IUseRenderOptions): IUseRender => {
   const { site, activePage } = options
   const render = useRender(options)
 
+  const reusableStyle = computed(() => {
```
abemscac, line 22: I think the bug show up because it's meant to be flattened based on the active breakpoint, like how component custom style did, instead of being converted to media query css.
I think there's a chance that we could do this with [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_container_queries), which might make the code a little bit cleaner (or not 🤔?), but I'm not sure if the [browser support](https://caniuse.com/css-container-queries for that feature is good enough to be considered common.

sampullman: Got it! I like the idea of using container queries, but this works fine for now. I'm not completely sure how to reproduce the original bug, but everything seems correct in the templates. I'll keep testing!

sampullman:
No problem, looks good to me!

PR #196, opened 2024-01-02
feat(site-api): warn user if editor is open in another window #937

ClaireMao1996:
Close https://github.com/samatechtw/pubstudio/issues/937

### DIFF
--- a/backend/site-api/src/app/site/update_site.rs
+++ b/backend/site-api/src/app/site/update_site.rs
@@ -55,7 +55,7 @@ pub async fn update_site(
     if let Some(value) = &dto.pages {
         validator.validate_pages(value)?;
     }
-
+    let has_update_key = dto.update_key.is_some();
     let site = context
         .site_repo
         .update_site(&id, dto)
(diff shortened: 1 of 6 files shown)


backend/site-api/src/db/site_repo.rs
```diff
@@ -186,21 +189,23 @@ impl SiteRepoTrait for SiteRepo {
...
-        query.push(
-            " WHERE id = (
-                    SELECT MAX(id)
-                    FROM site_versions
-                )",
-        );
+        if let Some(update_key) = req.update_key {
+            // Append the condition to check the updated_at
+            query.push(" AND updated_at = DATETIME(");
+            query.push_bind(update_key);
```
sampullman, line 197: I made a small adjustment here, even though we verify `update_key` in the DTO, it's good to use `push_bind` to avoid the possibility of SQL injection.

backend/site-api/src/app/site/update_site.rs
```diff
@@ -64,6 +64,15 @@ pub async fn update_site(
             DbError::NoUpdate => ApiError::bad_request()
                 .code(ApiErrorCode::NoUpdates)
                 .message("No updates were made for the request.".to_string()),
+            DbError::EntityNotFound() => {
+                if has_update_key {
+                    ApiError::bad_request()
+                        .code(ApiErrorCode::UpdateStale)
```
sampullman, line 70: I added an error code here, I think it will make it easier to implement the frontend part.

sampullman:
Seems like this strategy will work well! I made a few changes in a new commit, let me know if it makes sense.

If so, do you mind adjusting the platform API PR to use the error code as well? Meanwhile I'll work on the frontend implementation 🙏

ClaireMao1996:
Sure! No problem!

sampullman:
Thanks, I'll merge this now so it's easier to work with locally without adjusting the skaffold deployment.

PR #197, opened 2024-01-03
fix: tab to select next component

### LINKED ISSUE #158: Fix tab to select next component
Pressing the tab key when a component is selected in the editor should select the next component (or child component). Currently, only shift-tab is working.

When a text component is selected, but ProseMirror editor is not active, tab should also select the next component. Currently it activates the ProseMirror editor.

abemscac:
Close #158

### DIFF
--- a/libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
+++ b/libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
@@ -229,6 +229,9 @@ export const BuilderDndComponent = defineComponent({
         // TODO -- is this necessary?
         key: renderKey.value,
         ref: dnd.elementRef,
+        // Make builder dnd component not tab-focusable to avoid jumping between
+        // activeElement and component when on Tab press.
+        tabindex: '-1',
       } as Record<string, unknown>
 
       const tag = component.value.tag
--- a/libs/frontend/ui-build/src/components/component-menu/ComponentDimensions.vue
+++ b/libs/frontend/ui-build/src/components/component-menu/ComponentDimensions.vue
@@ -14,7 +14,7 @@
         <SizeInput
           :cssProp="Css.Width"
           class="size"
-          @update="setStyle(Css.Width, $event)"
+          @update="updateStyle(Css.Width, $event)"
         />
         <div class="size-label">
           {{ t('build.min') }}
@@ -22,7 +22,7 @@
         <SizeInput
           :cssProp="Css.MinWidth"
           class="size"
-          @update="setStyle(Css.MinWidth, $event)"
+          @update="updateStyle(Css.MinWidth, $event)"
         />
         <div class="size-label">
           {{ t('build.max') }}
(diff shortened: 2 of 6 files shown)


libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
```diff
@@ -229,6 +229,9 @@ export const BuilderDndComponent = defineComponent({
         // TODO -- is this necessary?
         key: renderKey.value,
         ref: dnd.elementRef,
+        // Make builder dnd component not tab-focusable to avoid jumping between
+        // activeElement and component when on Tab press.
+        tabindex: '-1',
```
abemscac, line 234: Hi @sampullman 

This attribute was added because an element will be focused on mouse click, and that seemed to interfere the tab navigation feature in the builder. Without this attribute the builder would jump between activeElement and the next component when Tab is pressed.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/6b60fdd9-aa2e-446f-a5fd-827d3ea1bdf1

libs/frontend/ui-build/src/components/component-menu/ComponentDimensions.vue
```diff
@@ -135,6 +135,22 @@ const toggleCollapse = () => {
   )
 }
 
+const updateStyle = (property: Css, value: string | undefined) => {
```
abemscac, line 138: This is to fix an issue where jumping to the next size input by pressing Tab would directly update the style of the selected component.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/8262f6df-808b-43ad-898f-12ab216c779a

libs/frontend/ui-build/src/components/component-menu/SizeInput.vue
```diff
@@ -59,6 +61,13 @@ const emit = defineEmits<{
   (e: 'update', size: string | undefined): void
 }>()
 
+// Set tabindex to -1 when right menu is not focused to make size input not tab-focusable when
+// right menu is not focused. In the builder, we can select the next component by pressing Tab
+// continuously. However, if the next component is currently not in the viewport, after the
+// scrollToComponent operation, the subsequent Tab press will focus on the first size input
+// instead of selecting the next component.
+const tabIndex = computed(() => (runtimeContext.rightMenuFocused.value ? 1 : -1))
```
abemscac, line 69: There's a strange behavior I don't understand; if the next component is in the viewport, pressing Tab would correctly select & gain focus on the next component. But if the next component is not in the viewport, after the scrollToComponent operation, the subsequent Tab press will focus on the first size input (probably because inputs are tab-focusable?) instead of selecting the next component.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/a25709b5-d1a8-4657-a7d0-a362c9d028fa

Unfortunately, I didn't a clean way to solve this problem 😕. My current workaround is to only enable the tab navigation of size inputs when right menu is focused.

libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
sampullman: 👍

libs/frontend/ui-build/src/components/component-menu/SizeInput.vue
sampullman: I see! I guess that's just how the browser handles the tab key normally. Your fix seems to work well 🙏

sampullman:
Looks good to me!

PR #199, opened 2024-01-03
feat: frontend for multi-edit warning #937

sampullman:
@ClaireMao1996 This can be improved, but it's working for me with identity sites. I think the `main` site-builder image isn't updated for some reason, so it's failing to save (due to `update_key` not being recognized). I'm going to push this now to see if it's working in stg, but it would be great if you can test it in the morning!


ClaireMao1996:
Hi @sampullman , 

Is the problem you encountered similar to the one below?
截圖 2024-01-04 上午11 03 35

sampullman:
Yeah! I think it's caused by extra fields during DTO validation. Or maybe something else in the request is wrong?

ClaireMao1996:
I've noticed that the current regular expression `REGEX_DATE` doesn't match the provided date format `2023-12-14T07:41:21Z`. However, it does match date formats like `2023-11-23T03:16:40.549489Z`, I think I need to modify the expression to make it compatible with both formats.

sampullman:
Oh, I remember this problem, I think it's due to Javascript converting from `string -> Date -> string`. I wonder if there will be a problem since the date won't match, maybe it also needs to be solved on the frontend 🤔 

Can you try adjusting the regex, and see if there's still a problem?

sampullman:
Oh, actually I think I see the problem. In this case we aren't parsing in Javascript, just passing back the same date to update-site (received in get-site or update-site response). Normally our frontend passes the 2 digit millisecond value, which is how the API is set up to parse.

Maybe this is solved by using `JsDate` in the update/get site API response? It might format the date correctly, otherwise changing the REGEX could work. But it would be nice to keep the same date format in all the API calls.

PR #211, opened 2024-01-07
209 tooltip behind modal

### LINKED ISSUE #206: Style input loses focus
Style value input loses focus when StyleRow order changes. When typing in a value for a custom/override/mixin CSS property, if the order changes the input will lose focus.

This can most likely be solved by adding a `key` to each row that doesn't change when the value is updated.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/5b303035-cead-436f-81f2-9aa3e3d8f6de

### LINKED ISSUE #209: Tooltips shown behind behavior modal
Info/explainer tooltips are shown behind the Behavior modal.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/c2c5945d-9ce3-42b4-9a84-d3eafaf24277

sampullman:
Close #209 

Close #206

### DIFF
--- a/libs/frontend/feature-build/src/lib/style/use-reusable-style-menu.ts
+++ b/libs/frontend/feature-build/src/lib/style/use-reusable-style-menu.ts
@@ -135,18 +135,20 @@ export const useReusableStyleMenu = (): IUseStyleMenuFeature => {
       activeBreakpoint.value,
     )
 
-    return Object.entries(flattenedStyles).map(
-      ([css, source]) =>
-        ({
-          pseudoClass: currentPseudoClass.value,
-          property: css as Css,
-          value: source.value,
-          sourceType: source.sourceType,
-          sourceId: source.sourceId,
-          sourceBreakpointId: source.sourceBreakpointId,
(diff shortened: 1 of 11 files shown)


libs/frontend/feature-build/src/lib/style/use-reusable-style-menu.ts
```diff
@@ -135,18 +135,20 @@ export const useReusableStyleMenu = (): IUseStyleMenuFeature => {
...
-          property: css as Css,
-          value: source.value,
-          sourceType: source.sourceType,
-          sourceId: source.sourceId,
-          sourceBreakpointId: source.sourceBreakpointId,
-          inheritedFrom: getInheritedFrom(source),
-        }) as IInheritedStyleEntry,
-    )
+    return Object.entries(flattenedStyles)
+      .sort((a, b) => a[0].localeCompare(b[0]))
```
sampullman, line 139: @abemscac I think including the sort function here (similar to component/override styles) solves the problem, but I might have overlooked something.

sampullman:
The z-indexes are getting complicated, so hopefully this didn't break something else.

PR #213, opened 2024-01-09
164 avoid affecting layout with selection border

### LINKED ISSUE #164: Avoid affecting layout with selection border
Find a way for to avoid affecting the editor site layout when a component is selected.

It may be necessary to use an absolute positioned overlay, instead of adding the border directly to the component.

### LINKED ISSUE #205: Pseudo styles not shown as inherited
When switching to another pseudo class, styles inherited from the default pseudo class should show inheritance info instead of a remove button.

Site example:
[230107_pseudo-bug.json](https://github.com/samatechtw/pubstudio-builder/files/13853534/230107_pseudo-bug.json)

https://github.com/samatechtw/pubstudio-builder/assets/59676941/d9e61eec-d192-4a4d-abf8-aa3972c07438

### LINKED ISSUE #207: Svg absolute position bug
Selecting an absolute positioned SVG causes its position to show incorrectly.

Example site:
[240107_svg-bug.json](https://github.com/samatechtw/pubstudio-builder/files/13853549/240107_svg-bug.json)

https://github.com/samatechtw/pubstudio-builder/assets/59676941/76614900-1b96-402f-822b-10d82da6d10b

### LINKED ISSUE #208: Img absolute position bug
Hovering an absolute positioned image in the editor causes it to be positioned incorrectly, or possibly disappear.

Example site:
[240107_img-absolute-bug.json](https://github.com/samatechtw/pubstudio-builder/files/13853562/240107_img-absolute-bug.json)

https://github.com/samatechtw/pubstudio-builder/assets/59676941/75b7d188-961a-4f1b-898e-fa911fbe3511

abemscac:
Close #164 
Close #205 
Close #207 
Close #208

### DIFF
--- a/libs/frontend/feature-build/src/lib/style/use-reusable-style-menu.ts
+++ b/libs/frontend/feature-build/src/lib/style/use-reusable-style-menu.ts
@@ -4,6 +4,7 @@ import {
 } from '@pubstudio/frontend/feature-site-source'
 import { useSiteSource } from '@pubstudio/frontend/feature-site-store'
 import {
+  currentStyleType,
   editCommands,
   editingMixinData,
   editStyles,
@@ -133,6 +134,7 @@ export const useReusableStyleMenu = (): IUseStyleMenuFeature => {
       breakpointStylesWithSource,
       descSortedBreakpoints.value,
       activeBreakpoint.value,
+      currentStyleType.value === StyleType.Mixin,
     )
 
     return Object.entries(flattenedStyles)
@@ -155,6 +157,13 @@ export const useReusableStyleMenu = (): IUseStyleMenuFeature => {
     if (entry.sourceBreakpointId !== activeBreakpoint.value.id) {
       return t('style.inherited_breakpoint', {
         breakpoint: site.value.context.breakpoints[entry.sourceBreakpointId]?.name,
+      })
+    } else if (
+      entry.sourcePseudoClass !== undefined &&
+      entry.sourcePseudoClass !== currentPseudoClass.value
+    ) {
+      return t('style.inherited_pseudo_class', {
+        pseudoClass: entry.sourcePseudoClass,
       })
     } else {
       return undefined
--- a/libs/frontend/feature-build/src/lib/style/use-reusable-style-menu.ts
+++ b/libs/frontend/feature-build/src/lib/style/use-reusable-style-menu.ts
@@ -134,7 +134,7 @@ export const useReusableStyleMenu = (): IUseStyleMenuFeature => {
       breakpointStylesWithSource,
       descSortedBreakpoints.value,
       activeBreakpoint.value,
-      currentStyleType.value === StyleType.Mixin,
+      true,
     )
 
     return Object.entries(flattenedStyles)
--- a/libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
+++ b/libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
@@ -4,14 +4,9 @@ import {
   registerCustomEvents,
   removeListeners,
 } from '@pubstudio/frontend/feature-render'
-import {
-  activeBreakpoint,
-  descSortedBreakpoints,
-} from '@pubstudio/frontend/feature-site-source'
-import { findStyles } from '@pubstudio/frontend/util-component'
 import { RenderMode } from '@pubstudio/frontend/util-render'
 import { runtimeContext } from '@pubstudio/frontend/util-runtime'
-import { Css, IComponent, IRawStyle, ISite, Tag } from '@pubstudio/shared/type-site'
+import { IComponent, ISite, Tag } from '@pubstudio/shared/type-site'
 import { buildContentWindowInnerId } from '@pubstudio/frontend/feature-build'
 import {
   computed,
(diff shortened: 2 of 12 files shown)


libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
```diff
@@ -36,43 +31,50 @@ const renderImageHover = (
...
+  if (imgElement) {
+    const imgComputedStyle = getComputedStyle(imgElement)
+
+    position = imgComputedStyle.position
+    top = imgComputedStyle.top
+    right = imgComputedStyle.right
 
-  const { width: rectWidth = 0, height: rectHeight = 0 } =
-    imgElement?.getBoundingClientRect() ?? {}
+    if (position === 'absolute' && buildContentWindowInner) {
```
abemscac, line 53: Hi @sampullman 

The good news is the absolute position `` seems okay now. However, the bad news is `` will now shifts a little when selected (whether it's absolute position or not 😕). Hover on another component seems to put it back to the correct position, though.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/700cf6bb-c939-4653-aabb-917b31672b3b

Some more serious cases:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/4c1e5c34-50c0-4b78-9a1a-cdb1761d7175

From the video we can also see the new selection overlay is not always in the correct position. I'll leave a comment to describe related issues below.

libs/frontend/ui-build/src/components/BuildDndOverlay.vue
```diff
@@ -79,35 +76,150 @@ const overlayStyle = computed(() => {
...
+watch(
+  () => ({
+    selectedComponent: editor.value?.selectedComponent,
+    // Re-compute selection overlay width & height when active breakpoint changes.
+    activeBreakpoint: activeBreakpoint.value,
+  }),
+  async ({ selectedComponent }) => {
+    if (selectedComponent) {
+      // This `nextTick` is necessary for buildContentWindowInner and componentElement
+      await nextTick()
```
abemscac, line 121: Here the style for selection overlay is calculated with `ref`+`watch` instead of `computed`, because I noticed the width&height for the bounding rect of the selected component is sometimes incorrect when a component selected. The dimensions is also incorrect when a component is removed & recovered. For example:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/d1545b75-d059-449f-a982-8f8c243cf960

Using `watch` with `nextTick` here seems to fix the issue. However, the overlay size is still incorrect when the page is loaded:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/c509d1a6-3477-42b1-9746-127fbc5a00af

I wonder if using an interval to periodically update the overlay style would make things easier; but that sounds like a strange hack 😵!

libs/frontend/feature-build/src/lib/style/use-reusable-style-menu.ts
```diff
@@ -133,6 +134,7 @@ export const useReusableStyleMenu = (): IUseStyleMenuFeature => {
       breakpointStylesWithSource,
       descSortedBreakpoints.value,
       activeBreakpoint.value,
+      currentStyleType.value === StyleType.Mixin,
```
abemscac, line 137: For #205 I just added an optional arg to decide if `sourcePseudoClass` should be included in the returned value of `computeComponentFlattenedStyles`. I hope this is okay 😬.

libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
sampullman: I see! The first case isn't so bad, but the second is a little too obvious.

libs/frontend/ui-build/src/components/BuildDndOverlay.vue
```diff
@@ -79,35 +77,157 @@ const overlayStyle = computed(() => {
...
+  return {
+    left,
+    top,
+    width,
+    height,
+  }
+}
+
+onMounted(() => {
+  setTimeout(() => computeSelectionOverlay(editor.value?.selectedComponent), 1)
```
sampullman, line 229: I tried to solve the selected component border on load bug with this. It sort of works, but doesn't quite draw the border in the right place.

This is what the border normally looks like:
Screenshot 2024-01-10 at 12 16 07 AM

And then on load (after the fix):

Screenshot 2024-01-10 at 12 35 31 AM

I don't quite understand the code yet, so I'm not sure why it would be slightly more or less offset sometimes. I noticed that clicking to/from a component sometimes causes the offset to be different as well.

abemscac: Got it, I'll look deeper into this and try to fix it today.
Thank you so much for helping me out!

abemscac: Hi @sampullman 

Unfortunately, I still didn't manage to solve this problem 😕.
But I noticed that the selection overlay will cause the root component to shrink a little bit when selected:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/8c2da393-1409-4198-9192-2920e2ac6d55

Notice how the build content "grow" to the right when root component unselected, and shrink a little bit to the left when re-selected. Removing the selection overlay seems to fix this glitch; maybe this has something to do with the bug we're seeing, but I'm not 100% sure.

I think I'll go pick up some other issues so that I don't get stuck here for too long. Sorry about that! 😵

sampullman: Ok, I'll try to fix it when I get the chance. Meanwhile I think I'll cherry pick the other commits and merge them first!

sampullman:
Replaced with #219

PR #221, opened 2024-01-12
210 input values cleared in preview mode

### LINKED ISSUE #210: Input values cleared in preview mode
Input values entered by the user will be automatically cleared in preview mode due to periodic fetch.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/a4048f36-3119-450f-a374-38a9ebcef88b

### LINKED ISSUE #216: Some suggestions for editor
In the process of developing templates, I encountered some issues related to the editor:

1. When I input a string such as `$19.99 / month` it automatically converted to a link. I believe this happens because it contains `.` and `/` leading the system to interpret it as a link

https://github.com/samatechtw/pubstudio-builder/assets/101389820/bd281e1f-0d28-46a6-9f01-2c8907aaeb80

2. When I try to create a new behavior, the content within `helpers` is obstructed.

https://github.com/samatechtw/pubstudio-builder/assets/101389820/6dfbfe3f-8953-40e7-ab65-9654cd9de6d1

abemscac:
Close #210 
Close #216 

Hi @sampullman 

For #210, I didn't think of a good way to solve it. In addition to input values, styles added via behavior code are also cleared after a periodic fetch. I thought I could use the `key` attribute on VNode to preserve some elements during re-render, but unfortunately that didn't work out.

My current solution is to compare the old site and the new site, and only update the site when content that'll affect the rendered output is changed (so things like editor & history are excluded from the check). I worried that this solution may not be very performant, due to the fact that the check is run within every periodic fetch, but I didn't find another way to solve the page-states lost problem 😕.

### DIFF
--- a/libs/frontend/feature-site-store/src/lib/use-site-source.ts
+++ b/libs/frontend/feature-site-store/src/lib/use-site-source.ts
@@ -5,6 +5,8 @@ import {
   ISiteStore,
   SiteSaveState,
 } from '@pubstudio/shared/type-site'
+import { serializeSiteContext, serializePage } from '@pubstudio/frontend/util-site-store'
+import { store } from '@pubstudio/frontend/data-access-web-store'
 import { computed, ComputedRef, Ref, ref } from 'vue'
 
 // This file is here, instead of frontend/feature-site-source, to decouple the `site`
@@ -37,6 +39,11 @@ const isSaving = computed(() => {
 })
 
 export const useSiteSource = (): IUseSiteSource => {
+  // For scratch site.
+  let oldScratchSite = ''
+  // For API site.
+  let oldRestored: ISiteRestore | undefined = undefined
+
   const setRestoredSite = (restored: ISiteRestore | undefined) => {
     if (restored) {
       restoredSite = restored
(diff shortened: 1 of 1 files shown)


libs/frontend/util-edit-text/src/lib/text-to-link-keymap.ts
```diff
@@ -10,7 +10,7 @@ import {
 import { createLinkNode } from './prosemirror-utils'
 
 export const urlRegex = () =>
-  /(https?:\/\/(www\.)?)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\\+.~#?&/=]*)/g
+  /(https?:\/\/(www\.)?)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\.[a-zA-Z()]{1,6}\b([-a-zA-Z0-9()@:%_\\+.~#?&/=]*)/g
```
abemscac, line 13: This fix is for the "$19.99 being converted to link" issue described in #216.

libs/frontend/feature-site-store/src/lib/use-site-source.ts
```diff
@@ -60,14 +67,93 @@ export const useSiteSource = (): IUseSiteSource => {
...
+      }
+      oldScratchSite = currentScratchSite
+
+      const restored = await siteStore.value.restore(site.value.updated_at)
+      if (restored) {
+        setRestoredSite(restored)
+        console.log('Site updated:', site.value.updated_at)
+      }
+    }
+  }
```
abemscac, line 100: This is the code that replaces the site with the new one on preview page when the site is edited in another page.
I have to compare the raw state in `localStorage` because this is the only shared value I can get in two different browser tabs. Although we have the `LocalStoragePlugin` in `scratch-site-store.ts`, the state in the store will still not be updated without refreshing the page, given that they are two separate browser tabs.

sampullman:
That's ok! I'm going to separate the link fix and merge that first. Clearing the inputs isn't a critical problem, so I'll take some time later to review the code and think about whether there's an easier solution long term.

I wonder if we can add a "feature" to both the preview and live site that saves input content to localstorage by component ID, then restores it on page load. It seems like that could be convenient, but there are probably a lot of edge cases I haven't thought of.

sampullman:
URL fix moved to #223

abemscac:
Replaced by #245

PR #225, opened 2024-01-16
feat: ui for setting background image style

### LINKED ISSUE #160: UI for setting background image style
Add UI for setting a site asset as the `background` or `background-image` style of a component. We can follow a similar pattern as the `src` input for an `img` component.

For `background-image`, all we need to do is replace the style with `url('<site-asset-url>')`. The `background` property is slightly more complicated since we need to make sure to retain any non-url information. For example, replacing `background: no-repeat url('<site-asset-url>')` should overwrite the URL, but keep the `no-repeat`.

abemscac:
Close #160

### DIFF
--- a/libs/frontend/ui-build/src/components/StyleRow.vue
+++ b/libs/frontend/ui-build/src/components/StyleRow.vue
@@ -22,6 +22,7 @@
         ref="valueInputRef"
         :modelValue="style.value"
         class="value-input"
+        :class="{ small: showAssetButton }"
         :placeholder="t('value')"
         :isError="!!valueErrorMessage"
         @update:modelValue="updateValue"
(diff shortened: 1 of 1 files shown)


libs/frontend/ui-build/src/components/StyleRow.vue
```diff
@@ -173,6 +214,42 @@ const removeStyle = () => {
...
+}
+
+const replaceBackground = (assetUrl: string) => {
+  const cssValue = `url("${assetUrl}")`
+  const currentValue = style.value.value
+  if (CSS.supports('background', currentValue)) {
+    // Replace the background value and retain any non-url information.
+    // Find a more efficient & comprehensive way to extract attributes from CSS value.
+    const attributes = style.value.value
+      .split(/([^\s]+|"[^"]+"|'[^']+')/g)
```
abemscac, line 237: Hi @sampullman 

I found it difficult to correctly extracts all attributes from a valid `background` css value; as a developer I really hate these shorthand properties 😅. It took me a whole day to finally come up with a somewhat okay regex.

Splitting the value with a space couldn't work because spaces are allowed in some of the [constituent properties](https://developer.mozilla.org/en-US/docs/Web/CSS/background#constituent_properties) of `background`. The following is a list of the example `background` values on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/background):

- `green`
- `content-box radial-gradient(crimson, skyblue)`
  - `url("...")` cannot be used together with `radial-gradient(crimson, skyblue)`, so we have to find where it is and replace it.
- `no-repeat url("../../media/examples/lizard.png")`
- `left 5% / 15% 60% repeat-x url("../../media/examples/star.png")`
- `center / contain no-repeat url("../../media/examples/firefox-logo.svg")`

sampullman: I see! I think it would have been acceptable to only support the `background-image` property for this feature, but this seems to work well.

If there end up being problems we can extract the function and add some unit tests.

sampullman:
Looks good to me! The UI is getting a bit crowded, I tried to make some small improvements here and there, but may need to extract properties with extra UI, similar to how Webflow handles things.

PR #228, opened 2024-01-17
feat: font preview in theme menu

### LINKED ISSUE #151: Add font preview
When adding a font in the Theme menu, show a preview of the selected font. For Google fonts, we need to make sure the font is downloaded/available, since the code for doing that currently only works after the font has been added to the site.

In `libs/frontend/ui-build/src/components/theme-menu/ThemeFontEdit.vue`, there is some commented out code for including the preview (downloading Google fonts not implemented yet).

abemscac:
Close #151


libs/frontend/ui-build/src/components/theme-menu/GoogleFontSelect.vue
```diff
@@ -8,9 +8,16 @@
     :options="GoogleFontNames"
     :clearable="false"
     :allowAny="true"
+    :customLabel="true"
     @select="emit('update:modelValue', $event)"
     @click.stop
-  />
+  >
+    <template #default="{ label }">
+      <div :style="{ 'font-family': quotedFont(label) }">
```
abemscac, line 16: Hi @sampullman 

In addition to the preview text at the bottom, I also applied the corresponding font to the option label in the list.
But we probably don't want to download all of the fonts all at once just for the option label, so currently only the font that's previously selected will be displayed. I hope this is okay 😬.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/de97eb1d-2a7a-4270-9ce1-8dbbde5743e3

sampullman: I was thinking about the best way to handle this as well. I think your solution is good!

sampullman:
Thanks! While reviewing I noticed an interesting problem, if we update a font I would expect (as a user), all of the references to change to the new font. But in our case it just removes the old Google font and shows the default browser font. One way around this is to always uses theme variables so you only have to update the font in one place, but I wonder if there's a more convenient way to handle it.

One solution could be to add a "name" to the font that acts as a variable, so that replacing the font family just changes the variable reference. Or maybe there's a way to creatively make use of `@font-face`! I'll add an issue, we can try to address this later.

PR #229, opened 2024-01-17
feat: show checkbox for boolean args

### LINKED ISSUE #194: Show checkbox for boolean args
For event/input/behavior args, show a checkbox instead of text input when the arg type is `boolean`.

abemscac:
Close #194

### DIFF
--- a/libs/frontend/ui-build/src/components/component-menu/ComponentEventEdit.vue
+++ b/libs/frontend/ui-build/src/components/component-menu/ComponentEventEdit.vue
@@ -38,7 +38,18 @@
           :forceEdit="true"
           :immediateUpdate="true"
           @update="setEventParam(param[0], $event)"
-        />
+        >
+          <template v-if="param[1].type === ComponentArgPrimitive.Boolean" #input>
+            <Checkbox
+              class="event-param-checkbox"
+              :item="{
+                label: '',
+                checked: ['true', true].includes(param[1].value),
+              }"
+              @checked="setEventParam(param[0], $event.toString())"
+            />
+          </template>
+        </MenuRow>
       </div>
     </div>
     <!-- Event Behaviors -->
@@ -95,17 +106,18 @@
 <script lang="ts">
 import { computed } from 'vue'
 import {
+  ComponentArgPrimitive,
   ComponentEventType,
   ComponentEventTypeValues,
   IBehavior,
   IComponentEvent,
   IComponentEventBehavior,
 } from '@pubstudio/shared/type-site'
-import { Plus } from '@pubstudio/frontend/ui-widgets'
-import EditMenuTitle from '../EditMenuTitle.vue'
-import EventBehaviorRow from './EventBehaviorRow.vue'
+import { Checkbox, Plus } from '@pubstudio/frontend/ui-widgets'
 import { useBuild } from '@pubstudio/frontend/feature-build'
 import { setEditBehavior } from '@pubstudio/frontend/util-command'
+import EditMenuTitle from '../EditMenuTitle.vue'
+import EventBehaviorRow from './EventBehaviorRow.vue'
 import { IUpdateComponentArgPayload } from '../component-arg/i-update-component-arg-payload'
 import BehaviorModal from './BehaviorModal.vue'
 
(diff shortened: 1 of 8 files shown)


libs/frontend/ui-build/src/components/component-menu/ComponentEventEdit.vue
```diff
@@ -38,7 +38,18 @@
           :forceEdit="true"
           :immediateUpdate="true"
           @update="setEventParam(param[0], $event)"
-        />
+        >
+          <template v-if="param[1].type === ComponentArgPrimitive.Boolean" #input>
+            <Checkbox
```
abemscac, line 43: Show checkbox for boolean event params.
We don't have any boolean event params for now, but here's a preview of what it may look like:

Screenshot 2024-01-17 at 3 44 00 PM

libs/frontend/ui-build/src/components/component-menu/ComponentEventEdit.vue
```diff
@@ -207,22 +219,23 @@ const updateEventName = (name: ComponentEventType | undefined) => {
     if (!newEvent.value.eventParams?.interval) {
       newEvent.value.eventParams = {
         interval: {
+          type: ComponentArgPrimitive.Number,
           value: '1000',
         },
       }
     }
   } else if (name === ComponentEventType.ScrollIntoView) {
-    if (newEvent.value.eventParams?.interval === undefined) {
```
abemscac, line 215: I removed this if statement to make the behavior feel more natural when switching between different event types.
I think the if statement is here for preserving event params when event type changes, but that also prevents the params for scrollIntoView event to be added.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/72ce7d73-3d7a-46d5-bbc5-adc2071713e1

One way to preserve the event params for all event types individually might be to include the event type in the params, but that kind of clashes with our existing data structure and might take some effort to refactor.

libs/frontend/ui-build/src/components/component-menu/ComponentInputEdit.vue
```diff
@@ -17,7 +17,17 @@
       :immediateUpdate="true"
       :placeholder="t('value')"
       @update="newInput.type = $event as ComponentArgPrimitive"
-    />
+    >
+      <template #input>
+        <PSMultiselect
```
abemscac, line 22: Change the component-input type control from input to dropdown.

libs/frontend/ui-build/src/components/component-menu/ComponentInputs.vue
```diff
@@ -98,7 +98,9 @@ const inputArray = computed(() => {
     renderMode: RenderMode.Build,
     resolveTheme: false,
   })
-  return Object.entries(inputs)
+  return Object.entries(inputs).sort((entryA, entryB) =>
+    entryA[0].localeCompare(entryB[0]),
+  )
```
abemscac, line 103: The sorting was added to maintain the order of inputs while editing.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/18444dc1-073f-406b-80c4-fe0488640eb5

libs/shared/type-site/src/lib/i-component.ts
```diff
@@ -64,6 +64,7 @@ export type IComponentState = string | number | boolean | Array<unknown> | objec
 export type IEventParams = Record<string, IEventParamEntry>
 
 export type IEventParamEntry = {
+  type: ComponentArgType
```
abemscac, line 67: Event param type was added so that we can use it to display checkbox for boolean params.

sampullman: This makes sense, although technically we would need to migrate existing sites, since it's a new required field. One way around it is to make `type` optional, though that could be annoying.

In this case I don't think anything will break when `type` is empty, but in the future we can add a process for breaking changes to the data format (e.g. add code for migrating during/after deserialization, and increment the site format version)

libs/frontend/ui-build/src/components/component-menu/ComponentEventEdit.vue
sampullman: This change makes sense to me! I don't think preserving the event params is very important.

PR #242, opened 2024-01-19
fix: add error logs to s3_client #963

ClaireMao1996:
Related to https://github.com/samatechtw/pubstudio/issues/963

### DIFF
--- a/backend/lib-shared-site-api/src/clients/s3_client.rs
+++ b/backend/lib-shared-site-api/src/clients/s3_client.rs
@@ -64,18 +64,25 @@ impl S3Client {
         let expires_in = Duration::from_secs(600);
         let url = delete_object.sign(expires_in);
 
-        let response = reqwest::Client::new().delete(url).send().await;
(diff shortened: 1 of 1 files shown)


ClaireMao1996:
It seems that the backups were not successfully uploaded to S3, but the current error message is not clear enough. 
In this PR I added some logs to better observe the actual error.

截圖 2024-01-19 下午4 35 21

sampullman:
Got it, thanks! It's probably some environment or cloudflare configuration error.

ClaireMao1996:
The signature doesn't match!

截圖 2024-01-21 中午12 31 54

Maybe the access key that we provided in `s3_client.rs` is wrong?

sampullman:
The environment variable wasn't provided correctly in the CI build. I confirmed that the env is correct now, but I don't see a backup in R2 yet. I haven't had a chance to check if there are any more error logs.

ClaireMao1996:
Sorry, I didn't notice that you had already fixed it. I'll check it right now!

PR #245, opened 2024-01-20
fix: input values cleared in preview mode

### LINKED ISSUE #210: Input values cleared in preview mode

abemscac:
Close #210


sampullman:
@abemscac I merged this early so I could test in stg, but I'll leave some comments about the changes!

backend/lib-shared-types/src/entity/site_api/site_entity.rs
```diff
@@ -13,6 +13,7 @@ pub struct SiteEntity {
     pub pages: String,
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
+    pub content_updated_at: i64,
```
sampullman, line 16: I changed this to `i64` because it was getting complicated to make sure the data in the sqlite3 db matched the query update_key. It's also more efficient, and I don't there there are benefits to using a date string.

The reason we use a date/string for the previous `update_key` is that we already had `updated_at` in that format.

backend/site-api/db/sites/migrations/20240118113105_sites_add_content_updated_at.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE
+  site_versions
+ADD COLUMN
+  content_updated_at INTEGER DEFAULT 0;
```
sampullman, line 4: I changed it to `INTEGER`, but the original problem is that the migration was written with postgres syntax, which wasn't valid for sqlite3.

It's somewhat annoying to add a date column in sqlite3, you have to add the column with no default, and then fill the data with a separate command since sqlite3 doesn't allow `CURRENT_TIMESTAMP` in `ALTER TABLE` commands.

backend/site-api/src/app/site/update_site.rs
```diff
@@ -56,9 +57,23 @@ pub async fn update_site(
         validator.validate_pages(value)?;
     }
     let has_update_key = dto.update_key.is_some();
+
+    // Update `content_updated_at` if any of `defaults`, `context`, or `pages` changes.
+    let mut content_updated_at: Option<i64> = None;
+
+    if dto.defaults.is_some() || dto.context.is_some() || dto.pages.is_some() {
```
sampullman, line 64: Since I fixed the frontend to only send updated data, we can avoid the comparison here and just check if the fields are present in the DTO.

backend/site-api/src/db/site_repo.rs
```diff
@@ -219,19 +237,34 @@ impl SiteRepoTrait for SiteRepo {
...
+    async fn migrate_db(&self, pool: &SqlitePool) -> Result<(), DbError> {
+        // Migrate the database
+        sqlx::migrate!("db/sites/migrations")
+            .run(pool)
+            .await
+            .map_err(|e| DbError::Migrate(e.to_string()))?;
+        Ok(())
+    }
+
+    async fn migrate_all(&self, sites: Vec<SiteMetadataEntity>) -> Result<(), DbError> {
```
sampullman, line 258: I realized we didn't have a way to migrate each site DB, so I added some code for it. Now the migrations will be run when the app starts, and when a new site is created.

libs/frontend/feature-site-store-init/src/lib/use-api-store.ts
```diff
@@ -150,17 +150,21 @@ export const useApiStore = (props: IUseApiStoreProps): ISiteStore => {
   // The API timer is reset if currently active
   const save = async (site: ISite, options?: ISiteSaveOptions): Promise<void> => {
     const storedSite = storeSite(site)
-    store.site.setSite(storedSite)
     let changed = false
     for (const key in site) {
       const k = key as keyof IStoredSiteDirty
-      dirty.value[k] = site[k] !== store.site[k]?.value
+      dirty.value[k] = dirty.value[k] || storedSite[k] !== store.site[k]?.value
```
sampullman, line 156: The updated data was being compared before it was serialized, so it was always marked as dirty!

I also realized that if some data changed previously, we shouldn't overwrite the `dirty` value, otherwise it won't be send in the update.

abemscac: Got it!
Thanks for all the detailed explanation!

PR #250, opened 2024-01-23
fix: site backups not working in stg #963

ClaireMao1996:
Close https://github.com/samatechtw/pubstudio/issues/963

### DIFF
--- a/backend/lib-shared-site-api/src/clients/s3_client.rs
+++ b/backend/lib-shared-site-api/src/clients/s3_client.rs
@@ -118,6 +118,29 @@ impl S3Client {
         )
     }
 
+    pub fn presign_put_site_backup(
+        &self,
+        filename: &str,
+        expires: u64,
+        size: u64,
+    ) -> Result<Url, ApiError> {
+        // Try to convert size to i64
+        let size_i64: i64 = match size.try_into() {
+            Ok(value) => value,
+            Err(_) => {
+                return Err(ApiError::internal_error().message("Size is too large to fit into i64"));
+            }
+        };
+
+        self.presign_put(
+            &self.backup_bucket,
+            filename,
+            expires,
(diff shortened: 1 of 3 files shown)


backend/lib-shared-site-api/src/clients/s3_client.rs
```diff
@@ -118,6 +118,25 @@ impl S3Client {
         )
     }
 
+    pub fn presign_put_site_backup(
```
sampullman, line 121: Can we reuse the existing `presign_put` function? It seems like the only difference is we need to pass `self.backup_bucket` as a parameter.

sampullman:
Thanks! I just included one comment.

I think it's possible to just call `PutObject` directly without presign, but it's fine to reuse the existing presign method since it's already well tested.

ClaireMao1996:
I forgot to comment out the dev environment, so all my tests were successful locally previously😅. I just replaced the `s3_secret_access_key` with the one from the staging for testing. I found that using presign doesn't work; the `MissingContentLength` error still occurs. When I switch back to directly calling the PutObject method and additionally add the `header(header::CONTENT_LENGTH, content_size.to_string())`, another error appears: `connection error: unexpected end of file`

截圖 2024-01-23 下午1 48 05

sampullman:
Maybe the content_size calculation doesn't match the real file size? I think that could cause both the errors you're seeing.

ClaireMao1996:
The size calculation is to obtain the size of the .db file after performing VACUUM, I'll double check for that!

sampullman:
The changes make sense to me! Do you think it's ready to merge?

ClaireMao1996:
I think we can merge and see how it goes. I’ll keep tracing the logs tomorrow! Thanks🙏

PR #251, opened 2024-01-23
fix: prosemirror up/down skips blank lines

### LINKED ISSUE #172: ProseMirror up/down skips blank lines
Up and down keys skip blank lines in the ProseMirror editor.

https://github.com/samatechtw/pubstudio-builder/assets/550982/a937f8e3-6c88-4635-bb8e-147e49d4d7a4

abemscac:
Close #172 

Hi @sampullman 

I think it's finally solved! Turns out the bug is actually caused by one of the default browser behavior instead of ProseMirror.
The same issue can be observed in [this sandbox](https://stackblitz.com/edit/web-platform-egfkum?file=index.html). It seems that `p::after { content: '\200b'; }` with a `<br>` inside will be skipped when using keyboard navigation; the only way to set the cursor to these "empty" lines is clicking them with mouse.

If I remember correctly, `p::after { content: '\200b'; }` was added because empty `<p></p>` will collapse when ProseMirror editor is not active. So I modify the style of `.pm-p` here and there to make sure `::after { content: '\200b'; }` is only applied when the component is not selected (for builder), and when in Preview Page & web-site app.

### DIFF
--- a/libs/frontend/feature-preview/src/lib/preview-component.ts
+++ b/libs/frontend/feature-preview/src/lib/preview-component.ts
@@ -31,11 +31,11 @@ export const PreviewComponent = () => {
     setup(props: ILiveComponentProps) {
       const { site, component } = toRefs(props)
       const { custom } = computeEvents(site.value, component.value)
-      registerCustomEvents(component.value, custom, null, true)
+      registerCustomEvents(component.value, custom, null, false)
 
       onMounted(() => {
         const { custom } = computeEvents(site.value, component.value)
-        registerCustomEvents(component.value, custom, null, false)
+        registerCustomEvents(component.value, custom, null, true)
       })
       onUnmounted(() => {
         removeListeners(component.value)
--- a/libs/frontend/feature-template/project.json
+++ b/libs/frontend/feature-template/project.json
@@ -1,5 +1,5 @@
 {
-  "name": "frontend-feature-template",
+  "name": "frontend-feature-template-builder",
   "$schema": "../../../node_modules/nx/schemas/project-schema.json",
   "sourceRoot": "libs/frontend/feature-template/src",
   "projectType": "library",
--- a/libs/frontend/ui-theme/src/css/components/prosemirror-view.postcss
+++ b/libs/frontend/ui-theme/src/css/components/prosemirror-view.postcss
@@ -34,9 +34,6 @@
 .ProseMirror pre {
   white-space: pre-wrap;
 }
-.ProseMirror-trailingBreak {
-  display: none;
-}
 
 .code-wrap {
   width: 100%;
--- a/apps/builder-demo/src/app/App.vue
+++ b/apps/builder-demo/src/app/App.vue
@@ -44,12 +44,14 @@ onMounted(async () => {
   }
 }
 
-.pm-p {
-  &:after {
-    content: '\200b';
-  }
-  &.pm-p-placeholder {
-    cursor: text;
+.component-content-container {
+  .pm-p {
+    &::after {
+      content: '\200b';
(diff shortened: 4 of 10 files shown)


libs/frontend/feature-preview/src/lib/preview-component.ts
```diff
@@ -31,11 +31,11 @@ export const PreviewComponent = () => {
     setup(props: ILiveComponentProps) {
       const { site, component } = toRefs(props)
       const { custom } = computeEvents(site.value, component.value)
-      registerCustomEvents(component.value, custom, null, true)
+      registerCustomEvents(component.value, custom, null, false)
 
       onMounted(() => {
         const { custom } = computeEvents(site.value, component.value)
-        registerCustomEvents(component.value, custom, null, false)
+        registerCustomEvents(component.value, custom, null, true)
```
abemscac, line 38: I noticed that if we use `document.getElementById(componentId)` in the appear event of a component, it always return `null`. I think the `true` and `false` here should be swapped.

libs/frontend/feature-template/project.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "frontend-feature-template",
+  "name": "frontend-feature-template-builder",
```
abemscac, line 2: I added the `-builder` suffix to the `name` of some of the `project.json` in `pubstudio/pubstudio-builder` 
 because there are some project name conflicts between `pubstudio/libs` and `pubstudio/pubstudio-builder/libs`, otherwise Nx will throw an error when executing commands like `npx nx serve web`.

Screenshot 2024-01-23 at 3 44 59 PM

libs/frontend/feature-preview/src/lib/preview-component.ts
sampullman: Makes sense to me, but I wonder if it's a mistake or there was some other reason 🤔 
I'll keep an eye out!

libs/frontend/feature-template/project.json
sampullman: That should be fine! In theory we should also update the folder name and references, otherwise some NX commands might get confused. But in this case I think there's a low chance we'll need to use NX commands on the library, so this is a good solution.

libs/frontend/ui-theme/src/css/components/prosemirror-view.postcss
```diff
@@ -34,9 +34,6 @@
 .ProseMirror pre {
   white-space: pre-wrap;
 }
-.ProseMirror-trailingBreak {
```
sampullman, line 37: I think the reason why this was removed is that it results in extra spaces int the editor that don't exist in preview/live mode. You can see the difference when you press enter in the ProseMirror editor.

sampullman:
The explanation makes sense, but doesn't seem to be working as expected. Pressing the down arrow works to select the empty line, but pressing down again sends the cursor to the end of the editor:


https://github.com/samatechtw/pubstudio-builder/assets/550982/48f93def-44ff-45c5-a013-a2553e5e0168

abemscac:
Oh, sorry about that! It seems to work on my end 😬: 

https://github.com/samatechtw/pubstudio-builder/assets/59676941/cbdb8c9a-25f4-41c0-86ce-f6d0c5fe3168

I wonder if the behavior is observed because of some missing code in the `pubstudio` repo; there are some changes made to the `App.vue` inside that repository, so we have to checkout to `172-prosemirror-up-down-skips-blank-lines` in both repos to really fix the bug.

sampullman:
You're right! I thought I was on the correct branch, but must have switched before I tested.

PR #253, opened 2024-01-24
fix: site backups not working in stg #963

ClaireMao1996:


backend/lib-shared-site-api/src/clients/s3_client.rs
```diff
@@ -168,7 +161,12 @@ impl S3Client {
     pub async fn upload_backup(&self, object_key: &str, content_size: u64) -> Result<(), ApiError> {
         let url = self.presign_put_site_backup(&object_key, 600, content_size)?;
 
-        match reqwest::Client::new().put(url).send().await {
+        match reqwest::Client::new()
+            .put(url)
+            .header(header::CONTENT_LENGTH, content_size)
```
ClaireMao1996, line 166: After checking the new logs, I found that the `MissingContentLength` error still exists😥 It seems that it should be included in the actual HTTP request rather than in the signed URL. Therefore, I directly set the `Content-Length` in the `reqwest::Client::put request`. 
However, during testing, I encountered the same error message as yesterday: `connection error: unexpected end of file.` I am sure that the Content-Length is correct this time. 

After researching, this error could be related to network issues/server problems/authentication problems, etc. It's possible that there are issues in my local environment, so I'm considering merging it and testing it in the staging to confirm if the same issue persists. Alternatively, could you help test it, and are there more detailed error logs on the R2 server?

backend/lib-shared-site-api/src/clients/s3_client.rs
```diff
@@ -124,20 +125,12 @@ impl S3Client {
         expires: u64,
         size: u64,
     ) -> Result<Url, ApiError> {
-        // Try to convert size to i64
-        let size_i64: i64 = match size.try_into() {
```
ClaireMao1996, line 128: I removed this conversion to avoid the risk of errors.

backend/lib-shared-site-api/src/clients/s3_client.rs
sampullman: Unfortunately I can't see any logs in R2, only whether the object was added or not. Do you want me to merge this code now, or do you need to make more changes?

In order to make testing easier, should we add an endpoint to manually trigger a backup?

ClaireMao1996: Before merge, let me added an endpoint for testing.

ClaireMao1996: Oh we've already have `backup_all` endpoint, I think I can trigger it directly.

PR #254, opened 2024-01-24
fix: improve svg component editing #234

### LINKED ISSUE #234: Set SVG width/height with component style
Currently it's necessary to set the SVG width and height directly in the SVG modal. It should be possible to set the SVG width and height via component style.

This is likely due to styles not applied correctly to the wrapper in the editor renderer. Make sure it works in editor/preview/live mode!

sampullman:
Close #234


sampullman:
Improved SVG edit UI and added hint about setting width/height. It would be fairly complicated to make component styles override the SVG width/height, and probably not worth the effort at this point.

PR #258, opened 2024-01-25
236 mixin hover styles not applied in editor

### LINKED ISSUE #235: Color picker saturation not set correctly
When opening the background or text color dropdowns, sometimes the saturation is set correctly.

There also appear to be situations where the color is completely incorrect:

Screenshot 2024-01-18 at 2 12 04 PM

### LINKED ISSUE #236: Mixin hover styles not applied in editor
Pseudo styles set in mixins are not applied in the editor, when the active pseudo style changes.

For example, the hover background color is set to green here. It shows correctly in the background color style toolbar icon, but not in the editor:

Screenshot 2024-01-18 at 2 15 12 PM

### LINKED ISSUE #249: Selection border not updated
The selection border is not always updated as expected when styles or editor context changes, including when margin is updated in a reusable style, and when the size of a component changes between pseudo classes.

Making sure the border is recomputed on any site data change is probably enough to fix it.

https://github.com/samatechtw/pubstudio-builder/assets/550982/d89f8153-945e-4df4-a1aa-5fbddb427b8f

### LINKED ISSUE #252: List sites endpoint called repeatedly
When the site is updated in the editor, the list-sites endpoint is called repeatedly.

A component is most likely unmounted/mounted, and is calling list-sites in the setup or onMounted.

https://github.com/samatechtw/pubstudio-builder/assets/550982/a7c5e8f6-6bda-4912-9f60-234045abe549

abemscac:
Close #235
Close #236 
Close #249 
Close #252


libs/frontend/feature-render/src/lib/get-page-style.ts
```diff
@@ -51,10 +54,35 @@ export const getBuildPageStyle = (site: ISite, page: IPage): IRawStyleRecord =>
...
 
+    if (currentPseudoClass === CssPseudoClass.Placeholder) {
+      const resolvedPlaceholderStyle = rawStyleToResolvedStyleWithSource(
+        site.context,
+        placeholderStyles,
+      )
+      pageStyle[`#${component.id}${CssPseudoClass.Placeholder}`] =
+        resolvedPlaceholderStyle
+    }
+
```
abemscac, line 85: Hi @sampullman 

For #236 I wasn't able to reproduce the bug with the same template from staging; maybe it was fixed at some point in the past 🤔. However, I noticed that partial styles (inline styles applied to some of the content in a component) will not be effected by the selected pseudo-class, which makes sense to me, but might confuse some users in the future, and some of the style-related bugs might have something to do with this. For example:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/49ae3fb1-2664-4cf9-a4d5-5c4ab88928cb

In the video I set the color of the text to red first, then try to change the color from the right menu. Since the inline styles will always overwrite all other styles, users might think right menu isn't working if they forgot they've applied some partial styles to the same component before.

---

I also noticed placeholder styles are not shown in the build mode, so I made some changes here to make it visible when `::placeholder` pseudo-class is selected.
Originally I was intended to make placeholder styles visible in the builder when default pseudo-class is selected, but then I realized properties like `border`, `background-color`, etc are also available with `::placeholder`, in order to make the behavior more consistent, maybe it's better to only show placeholder styles when `::placeholder` pseudo-class is selected, just like how all other pseudo-classes do in the builder. What do you think?

https://github.com/samatechtw/pubstudio-builder/assets/59676941/48611633-4eb3-40a7-9a2e-d8379f354599

libs/frontend/feature-sites/src/lib/use-sites.ts
```diff
@@ -56,15 +56,16 @@ export interface ISitesFeature {
   validateSiteName: (name: string) => string | undefined
 }
 
+const loading = ref(false)
+const sites = ref<ISiteViewModel[]>()
```
abemscac, line 60: This change is for #252.
I changed some other places as well so that the `<CreateAssetModal>` in `Build.vue` will be the only place calling the `listSites` endpoint when the builder is mounted.
Since Build Page and other pages (Asset Manager, Site Dashboard) will never show up at the same time, I made these two properties into global refs for simplicity.

libs/frontend/ui-build/src/components/BuildDndOverlay.vue
```diff
@@ -159,14 +158,9 @@ const computeSelectionOverlay = async (selectedComponent: IComponent | undefined
...
-  () => ({
-    selectedComponent: editor.value?.selectedComponent,
-    // Re-compute selection overlay width & height when active breakpoint changes.
-    activeBreakpoint: activeBreakpoint.value,
-    // Recompute on window resize
-    buildContentWindowSize: runtimeContext.buildContentWindowSize.value,
-  }),
-  async ({ selectedComponent }) => {
+  site,
+  async (newSite) => {
```
abemscac, line 162: This is for #249

libs/frontend/util-builtin/src/lib/resolve-style.ts
```diff
@@ -86,7 +86,7 @@ export const resolveThemeVariables = (
         console.warn(`Variable not found: ${variableName}`)
       }
 
-      return value
+      return value?.trim()
```
abemscac, line 89: This is for #235.
It seems that values like ` ${color-text}` (there's a space at the beginning!) is also considered valid in Style Row. Those leading space(s) cause some of the functions in `pubstudio-builder/libs/frontend/feature-color-picker/src/lib/color-picker-util.ts` to produce the incorrect results.

libs/frontend/feature-render/src/lib/get-page-style.ts
sampullman: This all makes sense, but I am still able to reproduce the original bug. I can change any property in a reusable style with the `:hover` pseudo style set, and it doesn't show in the editor.

I just tested again, and it seems like the reusable (pseudo) styles are also not applied in preview/live mode.

sampullman:
@abemscac I rebased the branch on main, but feel free to override the branch if you have local changes!

libs/frontend/feature-render/src/lib/get-page-style.ts
abemscac: Oh, sorry about that! I didn't realize the bug is about reusable style until now 😵.
I can reproduce the bug on my end now. I'll try to fix it within this PR, thank you!

sampullman: Thanks 🙏

libs/frontend/feature-render-builder/src/lib/render-builder.ts
```diff
@@ -273,7 +276,7 @@ export const computePropsContent = (
     ...data.attrs,
     ...builderStyleProps?.builderProps,
     ...events.native,
-    class: data.mixins.concat(builderStyleProps?.builderClass ?? []),
+    class: data.mixins.concat(component.id, builderStyleProps?.builderClass ?? []),
```
abemscac, line 279: Hi @sampullman 

Sorry for the delay! Currently we use `#test-c-0` to apply component custom styles, and use `.reusable-style-0` to apply reusable styles. Since the `#id` selector has a higher weight than the `.class` selector in CSS, and we actually flattened those styles before using it to generate the `<style>` tag, component custom styles will always override reusables styles not only in the builder, but also in preview and live site.

So I changed selector from `#id:hover` to `.id._hover` (class-based pseudo class) in build mode so that component custom styles and reusable styles always have the same weight. If they have the same weight, component custom styles will always override reusable styles because `<ReusableStyle />` is placed/rendered before `ComponentStyle` in all 3 modes (build, preview, and live).

For preview & live sites, the only change being made was that `#id` now becomes `.id` due to the weight issue mentioned above. But `:hover` is still `:hover`; it is not converted to `.hover` in preview & live site. The class-based pseudo class is only visible in the builder because we always want to see the corresponding styles when active pseudo-class changes.

I think I didn't explain it very well, sorry about that! 😵. Feel free to let me know if you need further explanation!

sampullman: It makes sense to me, thanks for the effort it took to solve this one!

When I was last working on style rendering I think I rushed too much and didn't take time to think through all the precedences. I probably should have drawn a diagram first!

PR #276, opened 2024-01-28
fix: drag and drop image bugs #274

### LINKED ISSUE #274: Drag and drop upload image fails
Dragging and dropping an image into the editor results in an invalid URL as the image `src`

Screenshot 2024-01-26 at 3 04 10 PM

sampullman:
Close #274

### DIFF
--- a/libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
+++ b/libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
@@ -203,24 +203,6 @@ export const BuilderDndComponent = defineComponent({
         }
       }
 
-      if (
-        dndState.value?.hoverSelf ||
-        dndState.value?.hoverTop ||
-        dndState.value?.hoverRight ||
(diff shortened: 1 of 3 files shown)


libs/frontend/feature-render-builder/src/lib/dnd/builder-dnd-component.ts
```diff
@@ -203,24 +203,6 @@ export const BuilderDndComponent = defineComponent({
         }
       }
 
-      if (
```
sampullman, line 206: @abemscac I moved this to `use-drag-drop` because it seemed more efficient, and easier to follow what's going on with the code. I think there may be a situation where it causes the DND overlay to remain on screen when it shouldn't, but I can't reproduce it. That problem may be solved by simplifying further and moving the code that clears `runtimeContext.buildDndState` back to `use-drag-drop`.

abemscac: Got it, thanks for letting me know!

PR #277, opened 2024-01-28
fix: breakpoint styles incorrect in editor #273

### LINKED ISSUE #273: Breakpoint styles not displaying correctly in editor
Breakpoint, and possibly mixin styles, are not displaying correctly in the editor.

See this example site, selecting either the Mobile or Tablet(Small) breakpoints displays the site incorrectly, although preview/live works as expected:

[240126_blog-company-3.json](https://github.com/samatechtw/pubstudio-builder/files/14073245/240126_blog-company-3.json)

sampullman:
Close #273

### DIFF
--- a/libs/frontend/feature-render-builder/src/lib/compute-builder-reusable-styles.ts
+++ b/libs/frontend/feature-render-builder/src/lib/compute-builder-reusable-styles.ts
@@ -14,10 +14,10 @@ export const computeBuilderReusableStyles = (site: ISite): IRawStyleRecord => {
   const accumulatedBreakpointIds = new Set<string>([DEFAULT_BREAKPOINT_ID])
 
   for (const breakpoint of descSortedBreakpoints.value) {
+    accumulatedBreakpointIds.add(breakpoint.id)
     if (breakpoint.id === activeBreakpoint.value.id) {
       break
     }
-    accumulatedBreakpointIds.add(breakpoint.id)
   }
 
   // Compute styles in the current pseudo class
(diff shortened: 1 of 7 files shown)


libs/frontend/feature-render-builder/src/lib/compute-builder-reusable-styles.ts
```diff
@@ -14,10 +14,10 @@ export const computeBuilderReusableStyles = (site: ISite): IRawStyleRecord => {
   const accumulatedBreakpointIds = new Set<string>([DEFAULT_BREAKPOINT_ID])
 
   for (const breakpoint of descSortedBreakpoints.value) {
+    accumulatedBreakpointIds.add(breakpoint.id)
```
sampullman, line 17: @abemscac I had to move this line above the `activeBreakpoint`, otherwise the current breakpoint wouldn't be added.

libs/frontend/feature-render-builder/src/lib/compute-builder-reusable-styles.ts
```diff
@@ -28,13 +28,18 @@ export const computeBuilderReusableStyles = (site: ISite): IRawStyleRecord => {
     accumulatedBreakpointIds.forEach((breakpointId) => {
       const bpStyle = style.breakpoints[breakpointId]
       const rawStyle = bpStyle?.[CssPseudoClass.Default] ?? {}
-      result[`.${styleId}`] = { ...rawStyle }
+      const cls = `.${styleId}`
+      const currentStyle = result[cls] ?? {}
+      result[cls] = { ...currentStyle, ...rawStyle }
```
sampullman, line 33: @abemscac Instead of replacing the style, I think we need to merge. Otherwise styles from larger breakpoints aren't applied in the builder. This change was also necessary in `getBuildPageStyle`.

libs/frontend/feature-render-builder/src/lib/compute-builder-reusable-styles.ts
```diff
@@ -14,10 +14,10 @@ export const computeBuilderReusableStyles = (site: ISite): IRawStyleRecord => {
   const accumulatedBreakpointIds = new Set<string>([DEFAULT_BREAKPOINT_ID])
```
sampullman, line 14: I think we're relying on the ordering of the set, which is not good. It's probably better to make this an array, there shouldn't be any downside. Also, it can be moved out of the `appendStyle` for efficiency, since we only need to calculate the `accumulatedBreakpointIds` once per render.

I'll submit a separate PR with these changes to see if everything still works!

PR #295, opened 2024-02-01
fix: list-site-servers current_load incorrect #1000

ClaireMao1996:
Related to https://github.com/samatechtw/pubstudio/issues/1000


sampullman:
I moved the helper to `util-test` instead of `type-site`! Since `type-*` libraries should only have types/interfaces/enums. Otherwise looks good!

PR #303, opened 2024-02-06
fix: can't undo when mixin menu is open

abemscac:
[#982](https://github.com/samatechtw/pubstudio/issues/982)

Hi @sampullman 

I'll leave some comments to describe what I've changed later, sorry for the delay! 🙏


libs/frontend/util-command/src/lib/command.ts
```diff
@@ -17,6 +40,30 @@ const pushCommandHelper = (site: ISite, command: ICommand, clearRedo = true) =>
...
+  }
+  return false
+}
+
+const shouldCloseMixinMenuForUser = (newCmd: ICommand | undefined): boolean => {
+  const mixinCommands = [CommandType.SetMixinEntry, CommandType.EditStyleMixin]
+  if (
+    !newCmd ||
+    mixinCommands.includes(newCmd.type) ||
+    commandsInGroup(newCmd, mixinCommands)
```
sampullman, line 56: @abemscac I added these two lines, otherwise changing the name or style value in the mixin menu caused it to close! I hope I understood everything correctly.

sampullman:
Nice work! This is a complicated feature with a lot of edge cases so I'll keep testing, but will merge now since the basic functionality appears to work well.

One thing that can be improved is when the mixin menu is exited by selecting another component, a close-mixin-menu command is not issued. I created an issue for it ,since I think the feature is good enough for now and other bug fixes are more important! #304

libs/frontend/util-command/src/lib/command.ts
abemscac: Oh, I didn't notice this, sorry about that!
Thanks for the fix!

abemscac:
>One thing that can be improved is when the mixin menu is exited by selecting another component, a close-mixin-menu command is not issued.

Oh, you're right! I forgot about this one. 😬 For now a close-mixin-menu command will only be issued (automatically) when a new command is pushed to the history, if the mixin menu is currently opened and the new command is not meant to close it.

libs/frontend/util-command/src/lib/command.ts
```diff
@@ -8,6 +13,24 @@ import { undoCommand } from './undo-command'
...
+        action: UiAction.CloseMixinMenu,
+        params: {
+          mixinId: editingMixinData?.mixinId as string,
+          originComponentId: editingMixinData?.originComponentId,
+        },
+      },
+    }
+    applyCommand(site, closeMixinMenuCommand)
+    site.history.back.push(closeMixinMenuCommand)
+  }
```
abemscac, line 32: The purpose of these code is to push a close-mixin-menu command to the history for user, if the mixin menu is currently opened, and the new command is not meant to close the menu. For example, if a user adds a new component to the editor by clicking Text in the submenu when the mixin menu is opened, the mixin menu should be closed, and the new component should be selected.

Instead of finding out all cases where the same scenario would happen and handle them one by one, I thought it would easier to just add it here, right before we push a new command. It feels hacky, but I didn't think of a better solution. 😬 

In addition, as you mentioned, there are still some cases that's not being handled. We can try to fix them in #304!

libs/frontend/util-command/src/lib/edit-styles-data.ts
```diff
@@ -1,23 +1,19 @@
...
+import { setEditingMixinData } from './editor-helpers'
 
 // Names of editing style properties
 export const editStyles = ref(new Set<string>())
 // Command group for actively editing styles
 export const editCommands = ref<ICommandGroupData | undefined>()
 export const currentStyleType = ref<StyleType>()
 
-// Extra info for editing reusable styles/mixins
-export const editingMixinData = ref<ICloseMixinMenuData | undefined>()
```
abemscac, line 19: I moved `editingMixinData` from here to the editor context so that mixin menu state (opened/closed) can be restored after refresh.

libs/frontend/util-command/src/lib/update-ui/update-ui.ts
```diff
@@ -0,0 +1,22 @@
+import { IUpdateUiData, UiAction } from '@pubstudio/shared/type-command-data'
```
abemscac, line 1: In order to handle different update ui actions while supporting redo & undo, I found myself introducing this nested apply function pattern to the app. I wonder if there's a better solution 🤔?

sampullman: The types are a little odd, but I think it's fine!

PR #321, opened 2024-02-16
fix: prosemirror select drag and drop

### LINKED ISSUE #311: Prosemirror select/drag and drop
When a text component is selected and the prosemirror editor is enabled, dragging the cursor to create a selection sometimes causes the parent component to be dragged instead.

This seems to mainly occur immediately after the prosemirror editor is enabled, and possibly after scrolling down the page.

https://github.com/samatechtw/pubstudio-builder/assets/550982/ecd1100a-6b29-4c84-9486-3be18fa14642

abemscac:
Close #311


sampullman:
Nice, thanks! It might be slightly annoying that we can no longer easily drag a prosemirror component, but I can't think of a good solution. I think dragging from the component tree is ok though, and maybe we can add a hotkey (like cmd/ctrl) that allows dragging text components later.

PR #326, opened 2024-02-16
fix: override mixin property and undo #266

### LINKED ISSUE #266: Override mixin property and undo
Overriding a mixin property in a component, then undoing, doesn't work as expected. It sets the component property to the mixin property, but instead should return to the original state.

https://github.com/samatechtw/pubstudio-builder/assets/550982/033cd7fe-58c9-48be-a7de-938c383ff236

sampullman:
Close #266

### DIFF
--- a/libs/frontend/ui-build/src/components/component-menu/ComponentStyles.vue
+++ b/libs/frontend/ui-build/src/components/component-menu/ComponentStyles.vue
@@ -23,8 +23,8 @@
         :omitEditProperties="nonInheritedProperties"
         :error="!resolveThemeVariables(site.context, entry.value)"
         class="menu-row"
-        @setProperty="setProperty(entry, $event)"
-        @setValue="setValue(entry, $event)"
+        @setProperty="setPropertyWrap(entry, $event)"
+        @setValue="setValueWrap(entry, $event)"
         @edit="editStyle(entry.property)"
         @save="saveStyle(entry.property)"
         @remove="removeStyle(entry)"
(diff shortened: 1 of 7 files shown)


libs/frontend/ui-build/src/components/component-menu/ComponentStyles.vue
```diff
@@ -92,6 +94,27 @@ const addStyle = () => {
...
 
+const setValueWrap = (oldEntry: IInheritedStyleEntry, newValue: string) => {
+  // If the property was inherited, create a new entry so undo works as expected
+  // Otherwise, the style would be reverted to `oldEntry` on undo
+  if (oldEntry.sourceType !== StyleSourceType.Custom) {
+    createStyle({ prop: oldEntry.property, value: newValue })
+  } else {
+    setValue(oldEntry, newValue)
+  }
+}
```
sampullman, line 105: @abemscac This is my alternate solution, I haven't fully tested it but it seems to work. The idea is that if we correctly track style changes here by making sure the inherited style isn't used as `oldStyle` in the command, undo will work as expeced.

libs/frontend/feature-build/src/lib/use-padding-margin-edit.ts
```diff
@@ -120,6 +120,9 @@ const drag = (e: MouseEvent) => {
           dragging.value.startY = e.clientY
         }
         const replace = !firstMove
+        // TODO -- to avoid the issue where undo sets the style back to an inherited
+        // value instead of removing it, we need to maintain `oldStyle: undefined` when
+        // replacing the command
```
sampullman, line 125: Unfortunately I didn't have time to get this to work for the margin/padding editor UI. The code is older so it's not using the new style helpers, but should be possible to fix.

sampullman:
I think I fixed the problems with padding/margin and dimension inputs, so I'll merge now.

All this style code is so complicated now, and I've noticed a lot of functions that do almost the same thing. I'll consider adding some E2E/unit tests and trying to simplify the code, so that it's more maintainable.

PR #327, opened 2024-02-17
fix: google font link bug

abemscac:
[#1033](https://github.com/samatechtw/pubstudio/issues/1033)

Turns out the bug is observed because we forgot to filter theme fonts before converting them to Google Font `<link>` 😅. This results in both native fonts and Google fonts to be included in `<link href="...">`.

`https://fonts.googleapis.com/css2?family=Georgia` actually exists, that's why we didn't see a 404 error in the console in this case. Only the 400 font-weight is defined in that CSS file by default, which caused all other font weights to lose effect. So the font weight would be broken right after the Georgia native font is added to the site.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/35de038f-2476-4feb-a716-5ab5363aa75c


sampullman:
That's an interesting bug, thanks for the quick fix!

PR #338, opened 2024-02-19
feat: adjust update-site-metadata endpoint for migrations #1043

ClaireMao1996:
Related to https://github.com/samatechtw/pubstudio/issues/1043

### DIFF
--- a/apps/api-site-test/test/integration/update-site-metadata.spec.ts
+++ b/apps/api-site-test/test/integration/update-site-metadata.spec.ts
@@ -1,5 +1,7 @@
+import { SiteType } from '@pubstudio/shared/type-api-platform-site'
 import {
   IGetSiteDomainsApiResponse,
+  IListSitesApiResponse,
   IUpdateSiteMetadataApiRequest,
 } from '@pubstudio/shared/type-api-site-sites'
 import { SiteApiResetService } from '@pubstudio/shared/util-test-reset'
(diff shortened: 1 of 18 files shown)


backend/site-api/src/db/sites_metadata_repo.rs
```diff
@@ -137,7 +137,7 @@ impl SitesMetadataRepoTrait for SitesMetadataRepo {
         let (mut query, count) = append_comma(query, "disabled", dto.disabled, count);
 
         if count == 0 {
-            return Err(DbError::NoUpdate);
+            return Ok(());
```
ClaireMao1996, line 140: I changed it here to return OK instead of an error to avoid having too many complex conditions like dto.disabled.is_some() &&... in `update_site_metadata.rs`.

apps/api-site-test/test/integration/update-site-metadata.spec.ts
```diff
@@ -14,88 +16,106 @@ describe('Update Site Metadata', () => {
...
+      expect(getBody.domains).toEqual(payload.domains)
     })
 
-    // Verify anonymous can no longer access
-    await api.get('/api/sites/current').set('Host', 'www.myblog.org').expect(403, {
-      code: 'SiteDisabled',
-      message: 'Forbidden',
-      status: 403,
+    it('returns 200 status and message when update site type', async () => {
+      payload.site_type = SiteType.Paid1
```
sampullman, line 55: I changed this to use the enum, and noticed I forgot to migrate a type file from the private repo, so I added it!

backend/site-api/src/db/sites_metadata_repo.rs
ClaireMao1996: Oh, sorry for overlooking that. I'll make sure to avoid that next time!

sampullman: No problem! I think your version is better, the API is not used that frequently so simpler/lower maintenance code is preferred.

PR #345, opened 2024-02-20
fix: component ellipses style not applied correctly

### LINKED ISSUE #237: Component ellipses style not applied correctly
The `text-overflow` style isn't applied as expected to text components, since there is a wrapper div around the content and text-overflow style isn't inherited by the child div.

One solution could be to manually apply the style to the child in the renderer.

It also may be useful to add specific UI for the text-overflow style, but that can be handled in a separate issue.

### LINKED ISSUE #331: Editing text color gradient fails
Setting a text color gradient and then editing the color manually in the component menu causes the gradient to show as the background.

Edit: It seems like this is only an issue with rendering in the builder - deleting and adding `-webkit-background-clip:
text` again fixes the issue.

Edit2: It's caused by the CSS property ordering, `-webkit-background-clip: text` should always be after `background`. Maybe we can sort the `-webkit` or `-` prefix to the end?

https://github.com/samatechtw/pubstudio-builder/assets/550982/9b163ba2-79ce-4937-9140-a7162cf03b7f

abemscac:
Close #237 
Close #331 

Hi @sampullman
 
I added a "Truncate" row below ComponentFlex in the component menu to control the ellipsis of a text-content component.

https://github.com/samatechtw/pubstudio-builder/assets/59676941/58b1ebae-4475-410d-8b10-ae37d9073b6f

One thing I'm not sure is that for nowrap ellipsis (one line ellipsis), it feels weird to only be able to see the first line of the text when ProseMirror editor is active. So currently the nowrap ellipsis will only be applied when the component is not selected (when the ProseMirror is inactive.) But this also means when users click on the "No Wrap" option, all text will still be visible until the component is deselected. I believe there's a better way to make this less confusing. 🤔


libs/frontend/feature-render/src/lib/get-page-style.ts
```diff
@@ -149,5 +151,34 @@ export const getLivePageStyle = (context: ISiteContext, page: IPage): IQueryStyl
...
+    result[selector] = sortRawStyle(rawStyle)
+    return result
+  }, {} as IRawStyleRecord)
+}
+
+// Order raw style by properties so that -webkit CSS properties would always
+// be added at the end. This is necessary in some cases, for example, for gradient
+// background color to work, `-webkit-background-clip: text` should always be after
+// `background`.
+const sortRawStyle = (rawStyle: IRawStyle): IRawStyle => {
```
abemscac, line 177: This is the fix for #331

abemscac:
By the way, I noticed that the gradient picker would overflow the menu after the recent UI changes.

Screenshot 2024-02-20 at 5 10 23 PM

Maybe we can have another issue to address this!

sampullman:
I'm going to move the fix for #331 to a separate branch and merge it first, there are a few things I want to consider about the ellipsis/truncate feature.

It works pretty much as I would expect, but it's a lot of complicated code for a feature which may not be used too often. I think we can still merge it, but may want to address a couple of things first:
- Pressing "right" when the prosemirror cursor is at the end of the truncated text is buggy, it shows overlapped text. This can be solved by not truncating text while editing - I think that makes sense from the UX perspective as well.
- I'm still not sure about deleting all related properties or not when un-selecting truncate. I think we can clear the styles when the "No wrap" button is clicked.
- UI can be clearer, people might not understand "No wrap", and be confused when they can't un-select it

It's not a big rush to solve these problems, and maybe we can consider alternate solutions. For example, a feature that allows filtering/transforming components before render would allow us to provide a "truncate" script that limits content by characters. I'm not sure if that's better or worse from the user perspective, but requires a lot of work 🤔 

Another odd thing is that `-webkit-box-orient` is not recommended in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient), although it seems to work fine in all major browsers.

abemscac:
Thanks! I think you're right, the UI is indeed confusing at the moment 😬, sorry about that!
As a temporary workaround I disable the truncation when the component is selected, and allow users to toggle the truncate style when the button is clicked.
I also change the label from "No Wrap / Line Clamp" to "Single Line / Multi Line", hoping to make it easier to understand.

>Another odd thing is that -webkit-box-orient is not recommended in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient), although it seems to work fine in all major browsers.

Yeah, I also saw that warning on the MDN docs, but it's surprising that they didn't offer any information about the alternative.😕
Maybe we should fallback to the filtering/transforming method you mentioned in the end.

sampullman:
Sounds good, I'll take a look!

My guess is that it was pretty complicated to implement and maintain for browsers, and not many people are using the feature since truncating with JS works well enough most of the time (and sometimes is preferred).

PR #356, opened 2024-02-21
fix: translations should not be wrapped with a div

### LINKED ISSUE #314: Translations shouldn't be wrapped with a div
Since translations are handled with the prosemirror editor, they're wrapped in a div. This makes it impossible to use translations for inline content.

Either set a specific class for the translation wrapper div and use CSS to make it inline, or strip the wrapper div.

Also, the translations modal preview should show the rendered text (without html tags):

i18n1

abemscac:
Close #314

### DIFF
--- a/libs/frontend/feature-site-translations/src/lib/use-site-translations.ts
+++ b/libs/frontend/feature-site-translations/src/lib/use-site-translations.ts
@@ -25,7 +25,7 @@ export interface ISiteTranslationsFeature {
   confirmDeleteTranslation: Ref<string | undefined>
   confirmSaveTranslationKey: Ref<string | undefined>
   confirmAddLanguage: () => void
-  selectActiveLanguage: (option: IMultiselectObj) => void
+  selectActiveLanguage: (option: IMultiselectObj | undefined) => void
   showEditTranslation: (key?: string) => Promise<void>
   saveTranslation: () => void
   deleteTranslation: () => void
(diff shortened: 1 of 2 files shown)


libs/frontend/feature-site-translations/src/lib/use-site-translations.ts
```diff
@@ -51,8 +52,17 @@ export const useSiteTranslations = (): ISiteTranslationsFeature => {
...
+      if (show && activeI18n && !(activeI18n in i18n)) {
+        // Set the active language back to default if it does not exist
+        // in the editor context. This is most likely to occur when the
+        // language is removed using the undo command.
+        const defaultLanguage = currentLanguages.value.find(
+          (language) => language.value === DEFAULT_LANGUAGE,
+        )
+        if (defaultLanguage) {
+          selectActiveLanguage(defaultLanguage)
+        }
```
abemscac, line 65: Hi @sampullman 

I noticed that the `lang` variable is not being used anywhere, so I removed it in this PR. Still want to double check if this make sense to you!

I also add some extra logic here to reset the active language back to default upon opening the TranslationModal, if the selected language does not exist anymore:

https://github.com/samatechtw/pubstudio-builder/assets/59676941/aa6651ed-47ee-4fb2-884d-b4aeba3664b0

abemscac:
Hi @sampullman 

Since the ProseMirror wrappers will only be removed after a save is triggered, I think we have to check all existing templates in staging & production to make sure all wrappers are removed as well.

sampullman:
Nice work, I think it's a clever solution! I tried to come up with some ways to make it fail, but could not.

I noticed the new language UI was a little odd - there's a checkmark even though clicking the language already adds it, so I changed it to an X.

PR #359, opened 2024-02-22
feat: admin endpoint for migrating site to another server #1046

ClaireMao1996:
Related to https://github.com/samatechtw/pubstudio/issues/1046


backend/site-api/src/app/site/create_site.rs
```diff
@@ -44,6 +46,11 @@ pub async fn create_site_helper(
         .await
         .map_err(|e| ApiError::internal_error().message(e))?;
 
+    if let Some(url) = backup_url {
```
ClaireMao1996, line 49: Due to the backup being on the old server before migration, there are no backup records available after creating the new site. If I call `/sites/:site_id/backups/:backup_id/actions/restore` on the new server, it will fail. 
I have considered copying the old backup record to the new server before calling restore, but I think this approach is not ideal. Instead, I adding `backup_url` to `create_site_dto` as an Option. If the value is `Some`, it indicates that the site to be created comes with default data. 
Let me know if this approach makes sense and if there are any improvements needed!

sampullman:
I added the admin UI and it seems to work well! But just before I pushed the changes I realized we also need to update the DNS entry for the site's subdomain. I included that but it will be hard to test, I'll create another site server in stg and try to migrate a few sites later.

PR #362, opened 2024-02-27
feat: memory cache for get-current-site requests #343

### LINKED ISSUE #343: Memory cache for get-current-site requests
Add a memory cache for get-current-site requests, which only includes content fields: `id, name, version, context, defaults, pages, created_at, updated_at`.

### Tasks
- Add memory cache for get-current-site data
- Reset cache (either clear it or include a "dirty" field) when `update-site` is called, for the currently published site-version
- Reset cache when a new site version is published

### LINKED ISSUE #344: Only include public pages in get-current-site
Parse the `pages` field and filter out pages with `public: false` in the get-current site endpoint.

This should be done before adding site data to the memory cache.

ClaireMao1996:
Close #343 
Close #344


ClaireMao1996:
Regarding the caching approach, I ended up splitting it into three caches: one for domain to site_id mapping, one for metadata, and one for site_usage. More tests need to be covered. I'll push the PR first in case you want to review/take a look. Let me know if there are any code that need to be improved. Thank you!

backend/site-api/src/app/site/get_current_site.rs
```diff
@@ -70,30 +72,60 @@ pub async fn get_current_site(
...
-    Ok(to_api_response(site))
+    // Only include public pages
+    let filtered_pages = filter_pages(&site.pages)?;
+    println!("Filtered Pages: {}", filtered_pages);
+
+    Ok(Json(to_api_response(&site, filtered_pages)))
+}
+
+fn filter_pages(pages: &str) -> Result<String, ApiError> {
+    let cleaned_pages = pages.trim().trim_matches('"').replace("\\", "");
```
ClaireMao1996, line 113: I got stuck here for a while because I encountered an error when converting the JSON string of pages into a `Map<String, Value>`. 

Initially, I hardcoded the exact same string and it passed. Then I converted both to bytes to compare them that I realized there might be hidden extra spaces, escape characters, and so on in the string obtained from `site.pages`. Therefore, I added this step to clean up the data.

backend/site-api/src/app/site/update_site_metadata.rs
```diff
@@ -40,11 +40,26 @@ pub async fn update_site_metadata(
             ApiError::internal_error().message(format!("Failed to save domains: {}", e))
         })?;
 
-    context
-        .metadata_repo
-        .update_site_metadata(&mut tx, &id, &dto)
-        .await
-        .map_err(|e| ApiError::internal_error().message(format!("Failed to update site: {}", e)))?;
+    if dto.disabled.is_some() || dto.site_type.is_some() {
```
ClaireMao1996, line 43: Revert to the original approach to avoid database lookups.

backend/site-api/src/app/site/get_current_site.rs
sampullman: That's because the JSON is "double" stringified! It's wasteful, but I couldn't figure out how to get things working between frontend/backend any other way.

The solution is to parse to `Value` first, and then parse the result as `Map<String, Value>`

apps/api-site-test/test/integration/get-current-site.spec.ts
```diff
@@ -129,54 +129,7 @@ describe('Get Current Site', () => {
 
       const body: IGetSiteApiResponse = response.body
       expect(body.name).toEqual('Test Site 3')
-      expect(body.pages).toEqual(pagesAfter)
-    })
-
-    it('returns current when requester is admin site for subdomain', async () => {
```
sampullman, line 27: I removed auth from `get-current-site`, since we only use the endpoint for serving the live site. It simplifies the code a bit and makes it faster!

backend/lib-shared-site-api/src/cache/cache.rs
```diff
@@ -33,11 +36,35 @@ pub type SiteMetadataCache = Cache<String, SiteMetadata>; // key: site_id, value
...
+    for (key, value) in pages.iter_mut() {
+        if let Some(public) = value.get("public").and_then(Value::as_bool) {
+            if public {
+                filtered_json.insert(key.clone(), value.clone());
+            }
+        }
+    }
+    let pages_json = serde_json::to_string(&filtered_json)?;
+    let filtered_pages = serde_json::to_string(&pages_json)?;
+    Ok(GetCurrentSiteResponse {
```
sampullman, line 58: I changed the cache to store `GetCurrentSiteResponse`, that way we don't need to do any processing in the endpoint.

backend/site-api/src/app/admin/reset_all.rs
```diff
@@ -48,6 +48,10 @@ pub async fn reset_all(
     // Clear all caches
     context.cache.cache.invalidate_all();
     context.cache.sync().await;
+    context.cache.metadata_cache.invalidate_all();
+    context.cache.metadata_cache.run_pending_tasks().await;
+    context.cache.domain_cache.invalidate_all();
+    context.cache.domain_cache.run_pending_tasks().await;
```
sampullman, line 54: I noticed some tests failing periodically after my changes, I'm not sure if they failed before but this fixes it!

backend/site-api/src/app/site/create_site.rs
```diff
@@ -54,10 +53,6 @@ pub async fn create_site_helper(
         .cache
         .create_or_update_usage(&site_id, &site, site_type)
         .await;
-    context
```
sampullman: Because the metadata cache helper takes care of updating the cache, we don't have to worry about it here.

backend/site-api/src/db/cache_helpers.rs
```diff
@@ -0,0 +1,75 @@
...
+                .map_err(|e| ApiError::not_found().message(e))?;
+
+            Ok(SiteMetadata {
+                owner_id: meta.owner_id,
+                site_type: meta.site_type,
+                disabled: meta.disabled,
+            })
+        })
+        .await
+    /*
```
sampullman, line 53: I forgot to remove this code 😅, I'll do it in my next PR. This function gets metadata from the cache, or updates the cache from the database if it's missing. I think the site data cache could be implemented similarly, but it was working fine and is a little more complex, so I left it as-is for now.

One thing I'm wondering is whether it's a good idea to use `Arc` for the site_data cache, and/or separate it from the usage. The reason is that `moka` clones the data when retrieving it, so we're potentially using a lot of memory/CPU every time it's accessed. It's kind of hard to measure though!

backend/site-api/src/app/admin/reset_all.rs
ClaireMao1996: Oh, I think that's the reason! Sorry, I forgot to add the relevant code here.

PR #371, opened 2024-03-03
fix: publish site cache bug #368

### LINKED ISSUE #368: Publishing new site fails
Publishing a new site (without a draft) doesn't always work.

- Reproduce bug and add failing test case
- Determine cause and fix

sampullman:
Close #368

### DIFF
--- a/backend/lib-shared-site-api/src/cache/cache.rs
+++ b/backend/lib-shared-site-api/src/cache/cache.rs
@@ -7,7 +7,7 @@ use lib_shared_types::{
 };
 use moka::future::Cache;
 use serde::Serialize;
-use serde_json::{json, Map, Value};
+use serde_json::{Map, Value};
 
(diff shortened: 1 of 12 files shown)


backend/lib-shared-site-api/src/cache/cache.rs
```diff
@@ -263,4 +243,24 @@ impl AppCache {
     pub async fn remove_metadata(&self, site_id: &str) {
         self.metadata_cache.invalidate(site_id).await;
     }
+
+    // Site Data
+    pub async fn get_site_with(
```
sampullman, line 248: @ClaireMao1996 I ended up splitting the site data cache from usage. There were some bugs in publishing and creating drafts, and I think the code is a little easier to follow this way. Instead of updating the cache when the site data changes, we can just clear it and let `get_current_site` refresh it when needed.

PR #373, opened 2024-03-04
feat: disable backups when S3 env is not present #6

### LINKED ISSUE #6: Disable backups when S3 env is not present
Disable automated site backups when S3 env vars are not provided.

### Tasks
- When `S3_URL`, `S3_ACCESS_KEY_ID`, or `S3_SECRET_ACCESS_KEY` are not provided, don't trigger automatic backups
- Backup related endpoint should return error `404`
- Print a warning on startup if backups are disabled

ClaireMao1996:
Close #6


sampullman:
Looks good to me!

PR #379, opened 2024-03-05
feat: reusable components

abemscac:
WIP [#1035](https://github.com/samatechtw/pubstudio/issues/1035)

Hi @sampullman 

This PR is still work in progress; we can only add a reusable component based on a selected component at the moment (and undo). The implementation is extremely simplified, and a lot of code & UI needs to be clean up & polish.

Currently there's a "To Reusable Component" button at the bottom of component menu. It is only visible when the selected component is not an instance of a reusable component:

Screenshot 2024-03-05 at 8 29 58 PM

After pressing it, a reusable component will be added to the site context based on the data of the selected component. Clicking on the entry will add a reusable component instance to the editor:

Screenshot 2024-03-05 at 8 30 05 PM

It would be great if we can make it easy to identify if a component is a "reusable instance" or not. For now I just add a new row in the component menu with some info:

Screenshot 2024-03-05 at 8 37 01 PM

Here's a list of TODO items related to this feature:

- [x] The content of a reusable instance should be editable (content override).
- [x] Only the `is` field of the inputs of a reusable instance is editable. Users cannot add or remove any input entry on a reusable instance.
- [x] The tag of a reusable instance cannot be edited.
- [ ] The style of a reusable instance should be uneditable until styling reusable component feature is supported.
- [ ] The children of a reusable instance cannot be moved or removed, and we cannot add components as the children of any reusable instance.
  - The intent is to make sure reusable instances always have the same children count as it's source.
- [ ] Component events and editor events for reusable components.
- [ ] Drag & drop feature for the items in the reusable component menu.
- [ ] Better UI for reusable component menu.
- [ ] UI for editing reusable components.
- [ ] UI for removing reusable components.
- What should we do if a user tries to remove a mixin (from the reusable style menu) which is referenced by a reusable component?

Overall I think my current code is quite messy. 😬
It would be great if you could take a look to see if I'm on the right track. Feel free to take over (or even start over!) if it'll make things easier for you, thank you!


sampullman:
@abemscac I'm taking a look at this now, but wanted to triple check to see if you have anymore local changes!

abemscac:
Hi @sampullman 

I don't have any local change, so it's ready to review this now. Thank you!

libs/frontend/data-access-command/src/lib/reusable-component/add-reusable-component.ts
```diff
@@ -0,0 +1,115 @@
...
+export const undoAddReusableComponent = (
+  site: ISite,
+  data: IAddReusableComponentData,
+) => {
+  const context = site.context
+  const { id, selectedComponentId } = data
+  const deleteCount = deleteReusableComponentWithId(site, id)
+  context.nextId -= deleteCount
+
+  if (selectedComponentId) {
```
sampullman, line 109: Do we still need this, since adding a reusable component doesn't set the selection?

libs/frontend/feature-render/src/lib/compute-attrs-inputs-mixins.ts
```diff
@@ -54,7 +54,8 @@ export const computeAttrsInputsMixins = (
   // Add attrs and inputs from default `inputs`
   for (const [key, input] of Object.entries(c.inputs ?? {})) {
     if (!inputs[key]) {
-      inputs[key] = resolveInput(input.is ?? input.default)
+      const overrideInput = c.reusableComponentData?.inputs?.[key]
+      inputs[key] = resolveInput(overrideInput ?? input.is ?? input.default)
```
sampullman, line 58: If we're letting users override the reusable component input, shouldn't this be:
```ts
inputs[key] = resolveInput(input.is ?? overrideInput ?? input.default)
```

libs/shared/type-site/src/lib/i-component.ts
```diff
@@ -96,6 +96,14 @@ export interface IComponentStyle {
   overrides?: IComponentStyleOverrides
 }
 
+export interface IReusableComponentData {
+  // Reusable component id
+  id: string
+  // Inputs override
+  // Key: input key, value: input `is`
+  inputs?: Record<string, unknown>
```
sampullman, line 104: Oh, I didn't notice this when I wrote one of my previous comments. Is there a reason for using this, instead of considering the component's existing `inputs` as overrides?

sampullman:
> What should we do if a user tries to remove a mixin (from the reusable style menu) which is referenced by a reusable component?

I suppose the same thing as removing a mixin referenced by a component? I see that we no longer show the mixin in the dropdown, but I think it's because the ID is invalid so it doesn't resolve to anything. Probably it's best this way, so that undo-ing a remove-mixin works as expected (e.g. the component gets the styles back). Otherwise it would be necessary to track everywhere the mixin was used in the remove-mixin command data.

Overall this looks like a great start! I will tidy up a few things, but won't make any major changes until thinking through the issue of whether to include the original reusable component in the page. The main purpose is to make editing the reusable component faster/more convenient, otherwise I don't think there's much benefit.

Some rough notes on how to achieve this:
- `reusableComponents` in the site context can just be a Set with component IDs
- No need for a special type of ID?

Actually I think that covers most of it, though I might be missing something 😓

sampullman:
I went ahead with making it so reusable components exist on the page, mostly because I think it makes the feature a lot easier to test for now. I still don't know if it's the best long term choice, but it's not too hard to switch back.

I made a little progress on overriding/inheriting style, but there's still a lot of work to do.

libs/frontend/feature-render/src/lib/compute-attrs-inputs-mixins.ts
```diff
@@ -29,11 +29,17 @@ export const computeAttrsInputsMixins = (
...
+  if (component.reusableSourceId) {
+    r = resolveComponent(context, c.reusableSourceId)
+  }
   const inputs: Record<string, unknown> = {}
   const attrs: Record<string, unknown> = {}
   const mixins: string[] = []
+  const content = c.content ?? r?.content
 
   if (editor) {
+    // TODO -- is this still used? pseudo styles still seem to work without it
```
sampullman, line 42: I noticed this while working on styles, it seems like we don't use the pseudo classes appended here anymore, but I'm not sure enough to remove the code yet.

libs/frontend/feature-render/src/lib/render.ts
```diff
@@ -28,12 +28,12 @@ export const computePropsContent = (
...
   const content: IContent = component.children?.length
     ? component.children.map((child) => renderComponent(site, child, renderMode))
-    : parseI18n(site, component.content)
+    : parseI18n(site, data.content)
 
   const props = {
     ...data.attrs,
     ...events.native,
-    class: data.mixins.concat(component.id),
+    class: data.mixins.concat(component.id, component.reusableSourceId ?? []),
```
sampullman, line 36: Rather than merge the styles (like how I tried to do it with component inheritance before), I'm just including the reusable component ID in the class list. I wonder if style precedence is always going to work 🤔

libs/shared/type-site/src/lib/i-component.ts
abemscac: Sorry for the late reply!
No, I think using the component's existing `inputs` as overrides is a lot better than my previous implementation. I'm not sure why I did that. 🤔

libs/frontend/feature-render-builder/src/lib/render-builder.ts
```diff
@@ -225,32 +230,54 @@ export const computePropsContent = (
 
   const { editor } = site
   const isSelected = editor?.selectedComponent?.id === component.id
+  const reusableCmp = resolveReusableComponent(
```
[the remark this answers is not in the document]
abemscac: Yeah, I think it makes sense!
I think there's also some data (mainly ids) that has to be modified when rendering a reusable instance. For example, the component id in `cmp.style.overrides` in a reusable component should be changed to the instance's children ids. We might have to do the same thing to the children ids in custom behaviors too, if there's any.

libs/frontend/feature-render/src/lib/compute-attrs-inputs-mixins.ts
abemscac: Yes, I think we still need this for non-default pseudo styles to work correctly due to how pseudo styles are added to `<style>` in the build mode (check [this comment](https://github.com/samatechtw/pubstudio-builder/pull/258#discussion_r1467651001) for more information).

https://github.com/samatechtw/pubstudio-builder/assets/59676941/b257891b-ae68-491b-8171-78c9f0637f86

libs/frontend/feature-render/src/lib/render.ts
abemscac: Yes, I think so!
From [this sandbox](https://stackblitz.com/edit/stackblitz-starters-vupjg7?file=index.html) it seems that the order of classes on an element will not effect the output (so `class="r g b"` and `class="g b r"` will lead to the same result). Instead, it is the order of styles defined in `<style>` that will effect the output, assuming that they all have the same priority (weight/specificity).

In `pubstudio-builder/libs/frontend/ui-build/src/components/BuildContent.vue:10` we render `<ReusableStyle />` first, then `<ComponentStyle />`, so component custom styles will always override reusable styles, unless the priority of styles defined in `<ReusableStyle />` is higher than that defined in `<ComponentStyle />`. It seems to work fine, at least for now. 😅

libs/shared/type-site/src/lib/i-component.ts
sampullman: Got it! There's still some work to do and details to figure out. I'll leave another comment about what I think the remaining tasks are.

libs/frontend/feature-render-builder/src/lib/render-builder.ts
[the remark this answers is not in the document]
sampullman: I'm not really sure how to handle child style overrides correctly. We want to be able to make style edits to the original/reusable component and have them show up in the instances. Since the IDs will be different, that will be pretty complicated and inefficient.

Maybe the best solution is to disable child override styles in reusable components? The main use case I've found so far for style overrides is to affect the hover style of children. It's occasionally useful but might not be a big problem to disable, and could always be added later if we come up with a clever implementation. Another option could be to only duplicate the overrides when a new instance is created, so the compromise is that changes to the reusable component style overrides aren't propagated.

libs/frontend/feature-render/src/lib/render.ts
sampullman: Right but I think we will run into trouble since the reusable component style will override the instance's style if it is rendered later (they're both in `ComponentStyle`). I think there are some straightforward ways to solve the problem though:
1. Since instances will always be created after the reusable component, and IDs are sequential, sorting them before render will ensure the instance style always overrides the reusable style. This seems a little hacky though.
2. Switch back to keeping reusable components separate, and render `ReusableComponentStyle` before `ComponentStyle`.
3. Keep components in the page, but change their IDs so they are more easily sorted/filtered. E.g. when a component is converted to reusable, change the ID from `test-c-1` to `test-rc-1` or similar.

I don't think it's necessary to solve this problem right away though, it's better to get everything else working first!

abemscac:
Replaced by #394

PR #380, opened 2024-03-06
fix: incorrect style precedence in editor

### LINKED ISSUE #378: Incorrect style precedence in editor
In the following situation, the editor displays the wrong style in the component menu, although the renderer is correct:

- Style is set in default breakpoint via component styles
- Style is set in lower priority breakpoint via mixin
- Renderer uses default breakpoint style, but editor shows mixin style

https://github.com/samatechtw/pubstudio-builder/assets/550982/ddef1c60-d81d-4b89-97b0-400457b7acea

abemscac:
Close #378

### DIFF
--- a/libs/frontend/util-component/src/lib/compute-component-flattened-styles.ts
+++ b/libs/frontend/util-component/src/lib/compute-component-flattened-styles.ts
@@ -7,9 +7,10 @@ import {
   IEditorContext,
   IRawStyleWithSource,
   IRawStylesWithSource,
+  StyleSourceType,
 } from '@pubstudio/shared/type-site'
 
-export const computeComponentFlattenedStyles = (
+export const computeFlattenedStyles = (
   editor: IEditorContext | undefined,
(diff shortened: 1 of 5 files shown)


libs/frontend/util-component/src/lib/compute-component-flattened-styles.ts
```diff
@@ -51,10 +54,19 @@ export const computeComponentFlattenedStyles = (
...
-        ;(result[css as Css] as IRawStyleWithSource).sourcePseudoClass =
-          currentPseudoClass
+      const key = css as Css
+      // Only update the result when the source type is component custom style
+      // since custom style has higher priority, or when the existing entry is
+      // not from component custom style.
+      if (
+        source.sourceType === StyleSourceType.Custom ||
+        result[key]?.sourceType !== StyleSourceType.Custom
+      ) {
```
abemscac, line 64: This seems to fix the issue. Custom styles always have the highest precedence than any other styles while flattening the styles.

PR #385, opened 2024-03-14
feat: builtin contact form #1086

sampullman:
Close 1086


libs/frontend/data-access-command/src/lib/component/add-component.ts
```diff
@@ -50,7 +50,22 @@ export const addComponentHelper = (site: ISite, data: IAddComponentData): ICompo
     parent.children.splice(parentIndex, 0, component)
   }
   context.components[id] = component
-  if (sourceComponent?.children) {
+  // Children overridden in builtin component definition take precedence
+  if (data.children) {
```
sampullman, line 54: @abemscac I added `children` to `IAddComponentData`, I couldn't find another way to include multiple levels of children in a builtin component.

For other components, we add specific IDs for each child, so that `resolveComponent` returns the expect builtin (when used as a child). But this doesn't work for a builtin with a complex structure that uses other builtin components as children.

There's probably a cleaner solution, but I hope this makes sense!

PR #389, opened 2024-03-18
1078 create table action

ClaireMao1996:
Close https://github.com/samatechtw/pubstudio/issues/1078


backend/site-api/src/app/admin/reset_all.rs
```diff
@@ -83,6 +89,13 @@ pub async fn reset_all(
                     },
                 )
                 .await?;
+
+                create_custom_table_helper(
+                    &context,
+                    &site_id,
+                    serde_json::from_str(SITE_SEED_CONTACT_TABLE).unwrap(),
```
ClaireMao1996, line 96: Currently, all three sites are seeded with the same table. If necessary, we can create different data for testing.

sampullman: I think it's fine for now 👍

sampullman:
I added a couple minor tests, otherwise this looks good!

backend/site-api/src/db/custom_data_repo.rs
```diff
@@ -40,35 +39,85 @@ impl CustomDataRepoTrait for CustomDataRepo {
...
+
+        for (column_name, column_info) in dto.columns.iter() {
+            let column_type = column_info.data_type.to_string();
+
+            query.push(format!(", {}", column_name));
+            query.push(" ");
+            query.push(column_type);
+            query.push(" NOT NULL");
+
+            for rule in column_info.validation_rules.iter() {
```
sampullman, line 59: This works, but we might eventually want to switch to validating in application code, instead of the database.

The reason is that since some validation must be implemented in the app, it's probably easier to maintain if we do everything that way. I think it also makes updating validation rules easier to reason about in some cases. That said, DB validation is probably more efficient, especially for `Unique`.

ClaireMao1996: Got it. So, should we keep these validations for now? Additionally, we'll also perform validations in the application code (if there are new validation rules added in the future, we'll add them to the application code).

sampullman: I think we can leave them in for now! I'll evaluate again before the feature is merged. After `ModifyColumn` is implemented, maybe it will be clearer whether the DB validations are too difficult to maintain.

PR #393, opened 2024-03-21
feat: add row action #1082 and list rows action #1084

ClaireMao1996:
Close https://github.com/samatechtw/pubstudio/issues/1082
Close https://github.com/samatechtw/pubstudio/issues/1084


apps/api-site-test/test/custom/add-row.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      payload.data = mockAddInvalidRow1()
+
+      return api
+        .post(testEndpoint(siteId))
+        .set('Authorization', adminAuth)
+        .send(payload)
+        .expect({
+          code: 'None',
+          message:
```
sampullman, line 91: We should try to return a specific error code here and custom message, so the frontend can show an appropriate error/warning.

It's a little annoying to do this with sqlx, but there are a few examples like the one in `backend/platform-api/src/db/repo/local_site_repo.rs` - `map_local_site_sqlx_err`

apps/api-site-test/test/custom/add-row.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      payload.data = mockAddInvalidRow2()
+
+      await api
+        .post(testEndpoint(siteId))
+        .set('Authorization', adminAuth)
+        .send(payload)
+        .expect({
+          code: 'None',
+          message:
+            'Query error: error returned from database: (code: 275) CHECK constraint failed: length(age) >= 1',
```
sampullman, line 107: Same comment about returning an error code for constraint violations. Another reason it's useful is if we change how constraints are implemented, it's only necessary to update the implementation, the code/message interface can stay the same.

apps/api-site-test/test/custom/add-row.spec.ts
ClaireMao1996: You're right, I will make the modification.

ClaireMao1996:
Hi @sampullman ,

This PR is ready for review again! Let me know if anything needs to be fixed.

backend/site-api/src/db/custom_data_repo.rs
```diff
@@ -27,6 +35,68 @@ pub struct CustomDataRepo {
...
+    Ok(row_data)
+}
+
+fn map_custom_data_sqlx_err(e: sqlx::Error) -> DbError {
+    let err_str = e.to_string();
+    match e {
+        sqlx::Error::Database(err) => {
+            let constraint_info = err.message().split(": ").collect::<Vec<&str>>();
+
+            if err.code() == Some(Borrowed("2067")) {
```
ClaireMao1996, line 71: The reason I checked with `err.code()` is because `err.constraint()` returns `None` in this case, so I use `code()` instead.

PR #395, opened 2024-03-25
feat: update row action #1083

ClaireMao1996:
Close https://github.com/samatechtw/pubstudio/issues/1083
Close https://github.com/samatechtw/pubstudio/issues/1079


ClaireMao1996:
Since it depends on #393, so I created the branch base on it. Once 393 is merged, we'll need to remember to rebase.

backend/lib-shared-site-api/src/db/util.rs
```diff
@@ -129,6 +135,63 @@ where
...
+            }
+            _ => {}
+        }
+
+        for rule in column_info.validation_rules.iter() {
+            match rule.rule_type {
+                RuleType::Unique => {
+                    query.push(" UNIQUE");
+                }
+                RuleType::MinLength => {
```
ClaireMao1996, line 172: When testing, I encountered an issue. If I add a column with a `MinLength` constraint, but because our default value is empty, this will cause a conflict. Perhaps, as you mentioned earlier, it would be better to execute validation conditions from the frontend? If we don't need the database to handle this, we can remove the code here.

sampullman: Thanks! Yes, I will consider moving validation to the application code. It seems convenient to have `UNIQUE` in the DB, but probably simpler overall to keep everything in the application.

